PR #102, opened 2022-04-12
Make clock suspending.

JayNewstrom:
This should support things like NTP as an implementation for the clock.


mikenachbaur-okta (approved):
LGTM. This won't be easy to do in Swift, so I might need to come up with an alternative approach...

PR #109, opened 2022-04-15
OKTA-467665 Migrate tokens from legacy OIDC SDK.

JayNewstrom:


mikenachbaur-okta (approved):
This all looks good, and is consistent with our previous discussion about the approach you planned to take.  I'm not overly fond of requiring the older SDK to be linked and included in the app in order to handle migration, but I suppose that might be difficult to work around.

PR #115, opened 2022-04-19
Don't allow CredentialBootstrap to be initialized multiple times.

JayNewstrom:


auth-foundation-bootstrap/src/main/java/com/okta/authfoundationbootstrap/CredentialBootstrap.kt
```diff
@@ -52,6 +52,9 @@ object CredentialBootstrap {
      * @param credentialDataSource the [CredentialDataSource] to associate with [CredentialBootstrap] as a singleton.
      */
     fun initialize(credentialDataSource: CredentialDataSource) {
+        if (privateCredentialDataSource != null) {
```
mikenachbaur-okta, line 55: Will this affect testing? And/or is this value reset between unit tests?

JayNewstrom: It definitely does! But I have an internal reset method to support that. https://github.com/okta/okta-mobile-kotlin/pull/115/files#diff-ae39fd7033b8f42e3154c0107c6037acb71c9b06a8f4e3f96070b5c7885640daL84

JayNewstrom: And called here: https://github.com/okta/okta-mobile-kotlin/pull/115/files#diff-406121440aad9f2fc4d56230287a50dd29cfecb5b435e25ee9a9b75098ecfe65R39

mikenachbaur-okta (approved):

PR #119, opened 2022-04-20
Move redirect from OidcConfiguration to call site.

JayNewstrom:
I'm not sure if I want to make this change, but opening for feedback.

### DIFF
--- a/oauth2/src/main/java/com/okta/oauth2/RedirectEndSessionFlow.kt
+++ b/oauth2/src/main/java/com/okta/oauth2/RedirectEndSessionFlow.kt
@@ -48,6 +48,7 @@ class RedirectEndSessionFlow private constructor(
          * The url which should be used to log the user out.
          */
         val url: HttpUrl,
+        internal val redirectUrl: String,
         internal val state: String,
     )
 
@@ -69,14 +70,15 @@ class RedirectEndSessionFlow private constructor(
      * Initiates the logout redirect flow.
      *
      * See [RedirectEndSessionFlow.resume] for completing the flow.
-     *
+     * @param redirectUrl the redirect URL.
      * @param idToken the token used to identify the session to log the user out of.
      */
-    suspend fun start(idToken: String): OidcClientResult<Context> {
-        return start(idToken, UUID.randomUUID().toString())
+    suspend fun start(redirectUrl: String, idToken: String): OidcClientResult<Context> {
+        return start(redirectUrl, idToken, UUID.randomUUID().toString())
     }
 
     internal suspend fun start(
+        redirectUrl: String,
         idToken: String,
         state: String,
     ): OidcClientResult<Context> {
@@ -84,10 +86,10 @@ class RedirectEndSessionFlow private constructor(
 
         val urlBuilder = endpoint.newBuilder()
         urlBuilder.addQueryParameter("id_token_hint", idToken)
-        urlBuilder.addQueryParameter("post_logout_redirect_uri", oidcClient.configuration.signOutRedirectUri)
+        urlBuilder.addQueryParameter("post_logout_redirect_uri", redirectUrl)
         urlBuilder.addQueryParameter("state", state)
 
-        return OidcClientResult.Success(Context(urlBuilder.build(), state))
+        return OidcClientResult.Success(Context(urlBuilder.build(), redirectUrl, state))
     }
 
     /**
@@ -97,7 +99,7 @@ class RedirectEndSessionFlow private constructor(
      * @param flowContext the [RedirectEndSessionFlow.Context] used internally to maintain state.
      */
     fun resume(uri: Uri, flowContext: Context): OidcClientResult<Unit> {
-        if (!uri.toString().startsWith(oidcClient.configuration.signOutRedirectUri)) {
+        if (!uri.toString().startsWith(flowContext.redirectUrl)) {
             return OidcClientResult.Error(RedirectSchemeMismatchException())
         }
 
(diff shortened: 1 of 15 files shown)


mikenachbaur-okta (commented):
Hmm I'm not sure; what's the benefit of relocating redirectUri?

I like the pattern you'd introduced, where the client is initialized with the clientId, scopes, and other parameters that uniquely identify this app within the admin dashboard.  If you want to authenticate to a different app, you create a different client.

The signInRedirectUri and signOutRedirectUri both are tied to the clientId, so I don't see how a developer would ever submit a different redirectUri for a given client.  Well, actually, I can (e.g. wildcard redirect URIs), but that's an edge-case and it's just as easy for a developer to create a different client.

JayNewstrom:
There are two things I was thinking about when I decided to give this a try:
1. Ensuring everything is initialized with what's required. Due to some flows not needing the redirect URLs. I've defaulted them as blank strings in the OidcConfig (in master), but that means you may forget to assign them if you're using a browser based flow, and it'll cause an error at runtime.
2. You may want to use different redirect URLs, (this may apply more to IDX, than WebAuthenticationUI), in which case maybe this isn't the right change. Maybe what I should do is make a param in the IDX SDK which defaults to the redirect URL, but can be specified elsewhere too.

mikenachbaur-okta:
I see what you mean. The approach I took was to make the common values (issuerURL, clientId, and scope) required properties on `OAuth2Client`.  However, in order to authenticate, you need to select a flow (e.g. AuthorizationCodeFlow, IDXAuthenticationFlow, etc).  Those flows that require additional values (like redirectUri) accept those values in those constructors.

So when the clientId is required, the code says something like `flow.client.configuration.clientId` to access that value.

JayNewstrom:
Yeah, this PR moves _closer_ to what swift is doing in that regard. I just chose the call site, rather than the constructor. But I'm not opposed to either. The biggest reason I chose here instead of the constructor is because you _could_ choose to only use login OR logout. And if it was specified in the constructor, you'd always need to supply _something_ even if it was a fake value.

oauth2/src/main/java/com/okta/oauth2/RedirectEndSessionFlow.kt
```diff
@@ -69,25 +70,26 @@ class RedirectEndSessionFlow private constructor(
      * Initiates the logout redirect flow.
      *
      * See [RedirectEndSessionFlow.resume] for completing the flow.
-     *
+     * @param redirectUrl the redirect URL.*
```
hansreichenbach-okta, line 73: unnecessary *?

hansreichenbach-okta:
> The biggest reason I chose here instead of the constructor is because you could choose to only use login OR logout. And if it was specified in the constructor, you'd always need to supply something even if it was a fake value.

I'm a fan of keeping required values close as possible to the flow that requires them for this reason. Nudging developers towards providing values they never use can lead to forgetful mistakes like redirecting to a dummy url and breaking a flow.

> I like the pattern you'd introduced, where the client is initialized with the clientId, scopes, and other parameters that uniquely identify this app within the admin dashboard. If you want to authenticate to a different app, you create a different client.

Maybe a bit off-topic but does this mean if you want 2 tokens from the same Auth Server with different scopes you'd need to make two different clients? It seems like this might be a non-zero use case for more sensitive apps where they get the sensitive scope every time but keep around a refresh token for the less sensitive scopes. Similar to transactional MFA use cases.

JayNewstrom:
> Maybe a bit off-topic but does this mean if you want 2 tokens from the same Auth Server with different scopes you'd need to make two different clients? It seems like this might be a non-zero use case for more sensitive apps where they get the sensitive scope every time but keep around a refresh token for the less sensitive scopes. Similar to transactional MFA use cases.

JayNewstrom:
I'm not opposed to going the direction of having a default, but also providing at the call site, but there are a few difference, and I'm not sure that's the best approach.

Scopes are required in `OidcConfiguration`, and defaulted at the call sites. 
Scopes are used in all flows, which is why it's required, and not defaulted.

redirectUrls _were_ defaulted to a blank string, to show they couldn't be null, but would fail at runtime if the developer didn't supply them when they went to do a browser based flow.

hansreichenbach-okta:
> Scopes are used in all flows, which is why it's required, and not defaulted.

Ah ok yeah so this is the primary difference. Always needed vs only needed based on scenario. Ok cool I like the new approach then

mikenachbaur-okta:
I'm late in this comment stream, but one of the solutions the Swift library takes for the unnecessary arguments problem (e.g. redirectUri and signoutUri) is to separate the login flow from the logout flow.  So for the `WebAuthentication` class, it takes a required signIn flow (`AuthorizationCodeFlow`), and an optional signOut flow (`SessionLogoutFlow`). Both flows consume an `OAuth2Client` (which is shared between both flows) which contains the common values for clientId and scope.  But the redirectUri values are assigned to the specific flow that requires it.

mikenachbaur-okta:
Regarding optional / defaulted values, I'd be worried that would add additional testing complexity, e.g. it might get confusing where a value was coming from, or we'd have to test the same feature in multiple scenarios (e.g. a value supplied as a default, or passed at the call-site).

And to the point about authenticating using different scopes, or different flows, in the Swift use-case the client is implicitly created, unless explicitly assigned by the developer. So for most use-cases a developer wouldn't really care if there was one, or several, client instances used at runtime.

mikenachbaur-okta (approved):

hansreichenbach-okta (approved):

PR #120, opened 2022-04-20
Emit an event when credentials are deleted.

JayNewstrom:


mikenachbaur-okta (approved):
This looks good. I'd like to see if we can get some sort of parity on the types of events that are broadcast between Swift and Kotlin. When credentials are deleted, added, when their tokens are replaced (e.g. in response to a refresh), etc.

JayNewstrom:
There's also a `TokenCreatedEvent` which will give you the `Credential` which has access to the old token/metadata, as well as the new `Token`. 

There's also a `CredentialCreatedEvent` which gives you the `Credential`.

So I _think_ we're in parity with the ones you've mentioned. But in general, I agree we should get parity here. This actually came up as I was reviewing one of your PRs :) I noticed I was missing this event.

PR #142, opened 2022-05-07
Add dependency verification.

JayNewstrom:
This is automatically checked by gradle for all tasks.


arvindkrishnakumar-okta (approved):

JayNewstrom:
This feels like it the right thing to do. But the tools aren't ready for this to not be a hassle for us going forward. I'm going to close. But I want this when the tools are ready!

PR #146, opened 2022-05-09
Prevent background starts when launching browser based sign in flows.

JayNewstrom:


JayNewstrom:
This is slightly more restrictive than https://developer.android.com/guide/components/activities/background-starts but I think it fits the usecase of this SDK nicely.

arvindkrishnakumar-okta (approved):

PR #177, opened 2022-06-28
OKTA-508186 Add support for Dokka publishing to github pages.

JayNewstrom:


mikenachbaur-okta (approved):

.github/workflows/deploy_docs.yml
```diff
@@ -0,0 +1,34 @@
+name: Deploy docs with Dokka
+
+on:
+  push:
+    tags:
+      - '*'
```
mikenachbaur-okta, line 6: I just noticed that you're only doing this on `tag`.  I'm not doing that (yet), but you might want to also trigger this on push to `master` so you can see a development version.  Like so:

https://github.com/okta/okta-mobile-swift/blob/master/.github/workflows/documentation-ghpages.yaml

JayNewstrom: I already pushed it to the gh-pages branch before I rebased and squashed my commits. So now it'll only be updated on releases, which was my intention. But happy to talk about other use cases too

PR #182, opened 2022-06-30
Add some defaults when deserializing.

JayNewstrom:
Found while testing with non Okta Authorization Servers.

### DIFF
--- a/auth-foundation/src/main/java/com/okta/authfoundation/jwt/Jwks.kt
+++ b/auth-foundation/src/main/java/com/okta/authfoundation/jwt/Jwks.kt
@@ -28,12 +28,12 @@ class Jwks internal constructor(
     internal val keys: List<Key>
 ) {
     internal class Key(
-        val algorithm: String,
+        val keyId: String,
+        val use: String,
+        val keyType: String,
+        val algorithm: String?,
         val exponent: String?,
         val modulus: String?,
-        val keyId: String,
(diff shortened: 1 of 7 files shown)


auth-foundation/src/main/java/com/okta/authfoundation/jwt/Jwks.kt
```diff
@@ -56,12 +56,12 @@ class Jwks internal constructor(
...
     @Serializable
     class Key(
-        @SerialName("alg") val algorithm: String,
-        @SerialName("e") val exponent: String?,
-        @SerialName("n") val modulus: String?,
         @SerialName("kid") val keyId: String,
-        @SerialName("kty") val keyType: String,
         @SerialName("use") val use: String,
+        @SerialName("kty") val keyType: String,
+        @SerialName("alg") val algorithm: String? = if (keyType == "RSA") "RS256" else null,
```
mikenachbaur-okta, line 62: Does this happen with some authorization servers? i.e. the `alg` is omitted?

mikenachbaur-okta (approved):

auth-foundation/src/main/java/com/okta/authfoundation/jwt/Jwks.kt
JayNewstrom: Yes, I've seen it in 2 now.

PR #190, opened 2022-07-13
Fix dokka by downgrading.

JayNewstrom:


.github/workflows/deploy_docs.yml
```diff
@@ -4,6 +4,8 @@ on:
   push:
     tags:
       - '*'
+  pull_request:
```
mikenachbaur-okta, line 7: Is this here for testing? Or do you need to include the target branch name the PR is merging to?

.github/workflows/deploy_docs.yml
```diff
@@ -27,7 +29,11 @@ jobs:
       - name: Build docs
         run: ./gradlew dokkaHtmlMultiModule
 
+      - name: Publish To Maven Local
+        run: ./gradlew publishToMavenLocal
+
       - name: Deploy to Github Pages
+        if: startsWith(github.push.ref, 'refs/tags/')
```
mikenachbaur-okta, line 36: FYI in my workflow, I trigger docs on pushes to master, and now on pushes to tags.  If there's no tag for the current commit, the docs are put into a `development` directory.

.github/workflows/deploy_docs.yml
JayNewstrom: Well, when I cut the release yesterday, publishing "javadoc" was broken (because we don't run it in CI). So now I'm running both forms of generating docs on every PR.

.github/workflows/deploy_docs.yml
JayNewstrom: Yeah, I saw that. I'm not doing any sort of versioning of docs right now. So only publishing the latest tag is my current intent. If we need to get fancier about it in the future, we can figure out requirements.

PR #191, opened 2022-07-15
Fix readme link.

JayNewstrom:


README.md
```diff
@@ -112,7 +112,7 @@ val oidcConfiguration = OidcConfiguration(
 )
 val client = OidcClient.createFromDiscoveryUrl(
     oidcConfiguration,
-    "https://{yourOktaOrg}.okta.com/.well-known/openid-configuration".toHttpUrl(),
+    "https://{yourOktaOrg}.okta.com/oauth2/default/.well-known/openid-configuration".toHttpUrl(),
```
arvindkrishnakumar-okta, line 115: The previous URL is still correct for Org AS case right?

JayNewstrom: While the previous URL works, it's not the same. And I'd like to encourage people to use the specific authorization server the intent.

For instance, my default authorization server has device_sso support, but when requesting the urls through the existing link, device_sso scope isn't supported. But when I update the URL to include the authorization server, device_sso is supported.

arvindkrishnakumar-okta (approved):

PR #201, opened 2022-07-25
OKTA-517853 Add revokeAllTokens to Credential.

JayNewstrom:


auth-foundation/src/main/java/com/okta/authfoundation/credential/Credential.kt
```diff
@@ -196,6 +200,37 @@ class Credential internal constructor(
...
+     */
+    suspend fun revokeAllTokens(): OidcClientResult<Unit> {
+        val localToken = token ?: return OidcClientResult.Error(IllegalStateException("No token."))
+        val pairsToRevoke = mutableMapOf<RevokeTokenType, String>()
+
+        pairsToRevoke[RevokeTokenType.ACCESS_TOKEN] = localToken.accessToken
+        localToken.refreshToken?.let { pairsToRevoke[RevokeTokenType.REFRESH_TOKEN] = it }
+        localToken.deviceSecret?.let { pairsToRevoke[RevokeTokenType.DEVICE_SECRET] = it }
+
+        return withContext(Dispatchers.Unconfined) {
```
rajdeepnanua-okta, line 216: What's the reasoning for using Dispatchers.Unconfined here?

JayNewstrom: I need to jump into coroutine land that allows me to do `async`, but I don't actually care about threading (the networking layer handles threading related to that already). As far as I know, that's the best way to accomplish it, but I'd be happy to make an updated if you know a better way :D

rajdeepnanua-okta (approved):

auth-foundation/src/main/java/com/okta/authfoundation/credential/Credential.kt
frett: 2 other options that would have worked here:

- `withContext(EmptyCoroutineContext) {}` - this would give you the `CoroutineScope` necessary to launch `async` without actually changing anything about the current coroutine context
- `coroutineScope {}` - this is probably the simplest solution, it creates a local coroutine scope that you can launch `async` tasks from

PR #209, opened 2022-09-09
Update dispatchers to be EmptyCoroutineDispatcher or use coroutineScope.

JayNewstrom:


JayNewstrom:
Update based off @frett suggestion [here](https://github.com/okta/okta-mobile-kotlin/pull/201#discussion_r967101756)

frett:
It seems weird that you are passing the `EmptyCoroutineContext` as a "dispatcher" in tests. a `Dispatcher` is just one type of object that can be stored within a  `CoroutineContext`, other types include the parent `Job` and an indicator object when using `coroutines-test` (context objects are extensible, and room leverages this to confine an active db transaction to a single thread).

the `withContext()` method allows you to replace any part of a coroutine context (not just the dispatcher), and it will merge those changes onto the inherited `coroutineContext`. The `EmptyCoroutineContext` object is just an empty object that will cause nothing to change when merging with the inherited context.

PR #211, opened 2022-09-09
Credential.tokenStateFlow

### LINKED ISSUE #210: Expose a Credential Token in a Coroutines StateFlow

Modern UI frameworks are built around observable data (`LiveData`, `StateFlow`, `State` in Jetpack Compose). Currently none of the objects within the SDK expose their data in an observable form. This means that external logic needs to be implemented to cause UIs to manually refresh their snapshot of data when state within the Okta SDK is updated.

This feature request is basically requesting that the `Credential` object exposes its current token via a `StateFlow` which would allow UIs to monitor the token for changes and automatically refresh their UI.

The reason I'm asking for access to the `token` in an observable format is to automatically drive a UI change when the user is logged in vs logged out.

While you are at it, it might be beneficial to also expose the `tags` via a `StateFlow` as well.

This should only affect the `Credential` object.

_No response_

### Additional Information?

I'd imagine this could be implemented via storing the private `_token` var as a `MutableStateFlow<Token?>`.
All access to it can then be accomplished internally using `_token.value`.
And a new property could be exposed: `val tokenStateFlow: StateFlow<Token?> get() = _token`

frett:
This exposes the current token for a `Credential` object as a `StateFlow` that can be observed.

Resolves #210

### DIFF
--- a/auth-foundation/src/main/java/com/okta/authfoundation/credential/Credential.kt
+++ b/auth-foundation/src/main/java/com/okta/authfoundation/credential/Credential.kt
@@ -30,8 +30,9 @@ import com.okta.authfoundation.util.CoalescingOrchestrator
 import kotlinx.coroutines.async
 import kotlinx.coroutines.awaitAll
 import kotlinx.coroutines.coroutineScope
+import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableStateFlow
-import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.transformWhile
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
 import okhttp3.Interceptor
@@ -58,7 +59,7 @@ class Credential internal constructor(
         keepDataInMemory = { false },
     )
 
-    private val _token = MutableStateFlow(token)
+    private val state = MutableStateFlow<CredentialState>(CredentialState.Data(token))
 
     @Volatile private var isDeleted: Boolean = false
 
@@ -73,15 +74,18 @@ class Credential internal constructor(
      */
     val token: Token?
         get() {
-            return _token.value
+            return state.value.token
         }
 
     /**
-     * A [StateFlow] that contains the current [Token] that's stored and associated with this [Credential].
+     * A [Flow] that emits the current [Token] that's stored and associated with this [Credential].
      */
-    val tokenStateFlow: StateFlow<Token?>
+    val tokenFlow: Flow<Token?>
         get() {
-            return _token
+            return state.transformWhile {
+                emit(it.token)
+                it !is CredentialState.Deleted
+            }
         }
 
     /**
@@ -136,22 +140,22 @@ class Credential internal constructor(
      * @param token the token to update this [Credential] with, defaults to the current [Token].
      * @param tags the map to associate with this [Credential], defaults to the current tags.
      */
-    suspend fun storeToken(token: Token? = _token.value, tags: Map<String, String> = _tags) {
+    suspend fun storeToken(token: Token? = this.token, tags: Map<String, String> = _tags) {
         if (isDeleted) {
             oidcClient.configuration.eventCoordinator.sendEvent(CredentialStoredAfterRemovedEvent(this))
             return
         }
         val tokenToStore = token?.copy(
             // Refresh Token isn't ALWAYS returned when refreshing.
-            refreshToken = token.refreshToken ?: _token.value?.refreshToken,
+            refreshToken = token.refreshToken ?: this.token?.refreshToken,
             // Device Secret isn't returned when refreshing.
-            deviceSecret = token.deviceSecret ?: _token.value?.deviceSecret,
+            deviceSecret = token.deviceSecret ?: this.token?.deviceSecret,
         )
         val tagsCopy = tags.toMap() // Making a defensive copy, so it's not modified outside our control.
         storage.replace(
             updatedEntry = TokenStorage.Entry(storageIdentifier, tokenToStore, tagsCopy),
         )
-        _token.value = tokenToStore
+        state.value = CredentialState.Data(tokenToStore)
         _tags = tagsCopy
     }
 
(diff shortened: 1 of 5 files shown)


JayNewstrom (commented):
You'll need a few things before this can be merged:
- Run `./gradlew apiDump` to get the latest public APIs updated 
- Add tests for the new `val`.
- The flow should be closed when the credential `delete` method is called.

JayNewstrom:
I guess StateFlow isn't closable. I'm not sure how to signal this when the credential is deleted. How do we signal through to the flow they need to look up a new credential?

frett:
Hmm, a StateFlow can't be closed according to the API docs. Their suggestion in the API docs is:
> All errors and completion signals should be explicitly materialized if needed.

Possible solutions I see:
- create a sealed wrapper object to reflect the state
- set the value to null when it's deleted, with no further updates to it (this would mirror the `token` property behavior of just returning null when the credential is deleted)
- expose a `Flow` instead of a `StateFlow`, which could be closed.

frett:
I think my preference would be expose it as a `Flow` that can be closed. Any preference from the API standpoint since it is your API to maintain long term ;)

JayNewstrom:
Yeah, I think I'd prefer to expose it as a `Flow`. A consumer can convert it to a `StateFlow`, using the `token` property, if they require it. But I'd like to expose it in a closable way.

frett:
I'll work up those changes now, I'm going to probably implement it as an internal sealed state hierarchy so that internal state stays consistent and atomic.

JayNewstrom:
One other thing, we'll also need you to sign the CLA https://developer.okta.com/cla/

auth-foundation/src/main/java/com/okta/authfoundation/credential/Credential.kt
```diff
@@ -69,7 +77,18 @@ class Credential internal constructor(
...
+        }
+
+    /**
+     * A [Flow] that emits the current [Token] that's stored and associated with this [Credential].
+     */
+    val tokenFlow: Flow<Token?>
+        get() {
+            return state.transformWhile {
+                emit(it.token)
+                it !is CredentialState.Deleted
```
frett, line 90: this will emit a final null value before closing the `Flow`, I can rework this a bit if we don't want that final null emitted.

frett: at least that is my understanding of how `transformWhile` works, this is my first time using it.

frett:
> One other thing, we'll also need you to sign the CLA https://developer.okta.com/cla/

just emailed the CLA. going to work on writing some tests now

frett:
@JayNewstrom thoughts on using the turbine library to test out the flow behavior?

https://github.com/cashapp/turbine

JayNewstrom:
Go for it 👍

auth-foundation/src/test/java/com/okta/authfoundation/credential/CredentialTest.kt
```diff
@@ -126,6 +144,15 @@ class CredentialTest {
         assertThat(removedEvent.credential).isEqualTo(credential)
     }
 
+    @Test fun testGetTokenFlowEmitsSingleNullAfterRemove(): Unit = runBlocking {
+        val storage = mock<TokenStorage>()
+        val credentialDataSource = mock<CredentialDataSource>()
+        val token = createToken()
+        val credential = oktaRule.createCredential(token = token, tokenStorage = storage, credentialDataSource = credentialDataSource)
+        credential.delete()
+        assertThat(credential.getTokenFlow().single()).isNull()
```
JayNewstrom, line 153: This case seems like it should instantly be closed, and not return any items.

JayNewstrom (commented):
Overall this looks good. Thanks for putting in the work for this.

auth-foundation/src/test/java/com/okta/authfoundation/credential/CredentialTest.kt
frett: k, I do think it should emit a final `null` value for anyone already observing a token flow.

I'll see what I can come up with to return an empty flow if the credential was already deleted without changing the final `null` behavior

JayNewstrom:
@frett I typically do a `./gradlew apiDump spotlessApply testDebugUnitTest` to make sure everything is good locally. (Looks like lint is failing right now).

auth-foundation/src/main/java/com/okta/authfoundation/credential/Credential.kt
```diff
@@ -94,10 +96,13 @@ class Credential internal constructor(
...
-            emit(it.token)
-            it !is CredentialState.Deleted
-        }
+        return state
+            .transformWhile {
+                emit(it)
+                it !is CredentialState.Deleted
+            }
+            .dropWhile { it !is CredentialState.Data }
+            .map { it.token }
```
frett, line 105: @JayNewstrom I ended up going with this instead of a simple `if (isDeleted) return emptyFlow()` because this avoids a minor race condition if the credential is deleted after that block before the actual flow is returned.

It was definitely an edge case that would give a flow with a single null before closing.

frett: I don't feel that strongly about accidentally emitting a single element flow of `null`, so I can switch it back if you would prefer the simplicity of that solution over a solution without the race condition

frett:
> @frett I typically do a `./gradlew apiDump spotlessApply testDebugUnitTest` to make sure everything is good locally. (Looks like lint is failing right now).

was just an unused import, just force pushed the fixup

JayNewstrom:
Everything looks good. Thanks again for the PR!

I'm going to take the weekend to mull over this to make sure I'm not missing anything, and I also need to wait on confirmation of the CLA from an internal team member.

frett:
Sounds good, do you have a rough timeline for when you will be publishing a new version? or is there a SNAPSHOT build published anywhere I can consume until a new release comes out? (obviously the snapshot won't include this until it gets merged)

JayNewstrom:
I'm hoping to get a released published next week. But that's not a promise! One of my coworkers is working on another feature we're hoping to land in 1.1!

We don't publish snapshots right now. You can always use something like gradle composite builds to pull it in locally to do development against, but that's likely not going to be worth setting up in your CI to be able to merge with it.

JayNewstrom (approved):
Thanks again for the PR!

PR #212, opened 2022-09-09
Introduce CredentialStoredEvent.

JayNewstrom:


auth-foundation/src/main/java/com/okta/authfoundation/credential/events/CredentialStoredEvent.kt
```diff
@@ -0,0 +1,40 @@
...
+ * limitations under the License.
+ */
+package com.okta.authfoundation.credential.events
+
+import com.okta.authfoundation.credential.Credential
+import com.okta.authfoundation.credential.Token
+import com.okta.authfoundation.events.EventHandler
+
+/**
+ * Emitted via [EventHandler.onEvent] when a [Credential.storeToken] was invoked after a [Credential.delete] call.
```
rajdeepnanua-okta, line 23: What is ```CredentialStoredAfterRemovedEvent``` for then?

JayNewstrom: This was a copy/pasta from that event 🙈 I'll fix the docs.

Good catch!

rajdeepnanua-okta (approved):

PR #214, opened 2022-09-11
Expose an errorIdentifier for IdTokenValidator.

### LINKED ISSUE #178: Explicit types/codes for errors

The old SDK uses explicit types for errors. For our use case, we map these errors to error codes so we can corroborate user reports with our crash logs. We don't use the error message itself to not expose information that could be used to aid a malicious user.

This new SDK is using a generic Error types with strings, which makes this approach brittle since strings can be changed without any obvious sign.

It would be great if the errors could be distinguished by type or by codes.

No idea

_No response_

### Additional Information?

_No response_

JayNewstrom:
Closes #178


JayNewstrom:
@idomo1 does this help with your use case?

rajdeepnanua-okta (approved):

idomo1:
@JayNewstrom Sorry for the late reply. Yes this is great, thanks!

PR #215, opened 2022-09-12
Add rate limit handling for network requests

rajdeepnanua-okta:


test-helpers/src/main/java/com/okta/testhelpers/NetworkDispatcher.kt
```diff
@@ -60,13 +60,9 @@ class NetworkDispatcher : Dispatcher() {
     }
 
     override fun dispatch(request: RecordedRequest): MockResponse {
-        var matchedEntry: Entry? = null
         val oktaRequest = OktaRecordedRequest(request)
-        enqueuedResponses.forEach { entry ->
-            if (entry.requestMatcher(oktaRequest)) {
-                matchedEntry = entry
-                return@forEach
-            }
```
rajdeepnanua-okta, line 69: This didn't seem to return the first matched request, but instead the last matched request. return@forEach didn't seem to stop forEach. So, I replaced .forEach with .first

test-helpers/src/main/java/com/okta/testhelpers/OktaRule.kt
```diff
@@ -54,11 +55,12 @@ class OktaRule(
     fun createConfiguration(
         okHttpClient: OkHttpClient = this.okHttpClient,
         cache: Cache = AuthFoundationDefaults.cache,
+        eventHandlers: List<EventHandler> = emptyList(),
     ) = OidcConfiguration(
         clientId = "unit_test_client_id",
         defaultScope = "openid email profile offline_access",
         okHttpClientFactory = { okHttpClient },
-        eventCoordinator = EventCoordinator(eventHandler),
+        eventCoordinator = EventCoordinator(listOf(eventHandler) + eventHandlers),
```
JayNewstrom, line 63: You might be able to avoid a lot of the dance (here and in individual tests) by using the `RecordingEventHandler.registerEventHandler`

auth-foundation/src/main/java/com/okta/authfoundation/client/internal/NetworkUtils.kt
```diff
@@ -117,6 +122,47 @@ internal suspend fun <T> OidcConfiguration.internalPerformRequest(
     }
 }
 
+internal suspend fun OidcConfiguration.internalGetResponseWithRateLimitHandlingFor(
+    request: Request
+): Response {
+    currentCoroutineContext().ensureActive()
+    val rateLimitErrorCode = 429
+
+     return withContext(ioDispatcher) {
```
JayNewstrom, line 131: You shouldn't need the `withContext`. All threading should be handled elsewhere already (I think).

auth-foundation/src/main/java/com/okta/authfoundation/client/internal/NetworkUtils.kt
```diff
@@ -117,6 +122,47 @@ internal suspend fun <T> OidcConfiguration.internalPerformRequest(
     }
 }
 
+internal suspend fun OidcConfiguration.internalGetResponseWithRateLimitHandlingFor(
```
JayNewstrom, line 125: I think we should rename this (and make it private?) Maybe just `executeRequest`? Open to ideas. But I don't think we need to have the rate limit name in it, that's an implementation detail.

auth-foundation/src/main/java/com/okta/authfoundation/credential/events/RateLimitExceededEvent.kt
```diff
@@ -0,0 +1,47 @@
...
+import okhttp3.Request
+import okhttp3.Response
+
+/**
+ * Emitted via [EventHandler.onEvent] when a network request receives 429 http status code (rate limit exceeded).
+ *
+ * The default implementation retries the request 3 times, with a minimum delay of 5 seconds between each request.
+ * See [maxRetries] and [minDelaySeconds]
+ */
+class RateLimitExceededEvent internal constructor(
```
JayNewstrom, line 28: You'll need to run `./gradlew apiDump` before lint will pass.

(PS, I normally do `./gradlew apiDump testDebugUnitTest spotlessApply` before committing).

auth-foundation/src/main/java/com/okta/authfoundation/credential/events/RateLimitExceededEvent.kt
```diff
@@ -0,0 +1,47 @@
...
+ * limitations under the License.
+ */
+package com.okta.authfoundation.credential.events
+
+import com.okta.authfoundation.events.EventHandler
+import okhttp3.Request
+import okhttp3.Response
+
+/**
+ * Emitted via [EventHandler.onEvent] when a network request receives 429 http status code (rate limit exceeded).
```
JayNewstrom, line 23:
```suggestion
 * Emitted via [EventHandler.onEvent] when a network request receives 429 HTTP status code (rate limit exceeded).
```

JayNewstrom, line 23: Can we link to the developer docs page from here as well?

test-helpers/src/main/java/com/okta/testhelpers/OktaRule.kt
rajdeepnanua-okta: Oh cool. Did not realize that existed

auth-foundation/src/main/java/com/okta/authfoundation/client/internal/NetworkUtils.kt
rajdeepnanua-okta: That does sound better, and captures exactly what it does. Will change the name to that

auth-foundation/src/main/java/com/okta/authfoundation/client/internal/NetworkUtils.kt
```diff
@@ -117,6 +122,45 @@ internal suspend fun <T> OidcConfiguration.internalPerformRequest(
...
+
+    if (response.code == rateLimitErrorCode) {
+        val event = RateLimitExceededEvent(request, response)
+        eventCoordinator.sendEvent(event)
+
+        for (i in 1..event.maxRetries) {
+            val timeToResetSeconds: Long
+            val requestId: String?
+            response.use {
+                val responseTimeSeconds = response.headers.getDate("Date")?.time?.milliseconds?.inWholeSeconds
```
JayNewstrom, line 141: Can you add a test where this is malformed? (for example, the date is `abcde`

auth-foundation/src/main/java/com/okta/authfoundation/client/internal/NetworkUtils.kt
```diff
@@ -117,6 +122,45 @@ internal suspend fun <T> OidcConfiguration.internalPerformRequest(
...
+                val retryTimeSeconds = response.header("X-Rate-Limit-Reset")?.toLong()
```
JayNewstrom, line 142: Can you add a test where this is malformed? For example the request id is `abcde`

auth-foundation/src/test/java/com/okta/authfoundation/client/NetworkUtilsTest.kt
```diff
@@ -354,6 +364,371 @@ class NetworkingTest {
...
         assertThat((result as OidcClientResult.Success<String>).result).isEqualTo("bar")
     }
+
+    @Test fun testPerformRequestWithNon429ErrorHasNoRetry(): Unit = runTest {
+        oktaRule.enqueue(path("/test")) { response ->
+            response.setResponseCode(401)
+            response.setBody("""{"foo":"bar"}""")
+        }
+        oktaRule.enqueue(path("/test")) { response ->
+            throw(IllegalStateException("Retried request after getting non-429 response."))
```
JayNewstrom, line 374: This is a super nit pick, but the syntax here is a little different than I'd expect.
```suggestion
            throw IllegalStateException("Retried request after getting non-429 response.")
```

Similar thing in other tests as well.

You can also remove the `response ->` from the line above in these.

rajdeepnanua-okta: Sure, that sounds good. Removing ```response ->``` gives me an override ambiguity error for some reason, so I'm leaving that in

JayNewstrom: You can make response `_` instead, which will remove the warning.

auth-foundation/src/main/java/com/okta/authfoundation/client/internal/NetworkUtils.kt
```diff
@@ -117,6 +122,45 @@ internal suspend fun <T> OidcConfiguration.internalPerformRequest(
...
+        val event = RateLimitExceededEvent(request, response)
+        eventCoordinator.sendEvent(event)
+
+        for (i in 1..event.maxRetries) {
+            val timeToResetSeconds: Long
+            val requestId: String?
+            response.use {
+                val responseTimeSeconds = response.headers.getDate("Date")?.time?.milliseconds?.inWholeSeconds
+                val retryTimeSeconds = response.header("X-Rate-Limit-Reset")?.toLongOrNull()
+                requestId = response.header("X-Okta-Request-Id")
```
JayNewstrom, line 143: What happens if multiple of any of these headers exist? `response.headers().values("foo")` returns a list (HTTP allows multiple response headers with the same key. One of your recent test additions sets multiple response headers, which made me think of this.

rajdeepnanua-okta: Good question. Going to look further into that. I didn't realize earlier that I was setting multiple response headers with the same key. I thought it was replacing the key, but didn't check the response in detail.

rajdeepnanua-okta: In this case, the value of the header appearing last in the response is taken. Also, the setHeader in OkHttp replaces the currently set value if the header already exists in the response, so I'm not testing that at the moment. Picking the last value in the response is something OkHttp is doing. So, is that something I should add a test for?

JayNewstrom: Sounds good. Thanks for looking into this! I'm fine without an explicit test for this 👍

JayNewstrom (approved):

PR #217, opened 2022-09-14
Skip retry for 429 requests with missing headers

rajdeepnanua-okta:


auth-foundation/src/main/java/com/okta/authfoundation/client/internal/NetworkUtils.kt
```diff
@@ -131,19 +131,20 @@ private suspend fun OidcConfiguration.executeRequest(
     var response = okHttpClient.newCall(request).await()
 
     if (response.code == rateLimitErrorCode) {
-        val event = RateLimitExceededEvent(request, response)
+        var event = RateLimitExceededEvent(request, response)
```
JayNewstrom, line 134: I thought we were doing this for very request before! Whoops!

auth-foundation/src/main/java/com/okta/authfoundation/client/internal/NetworkUtils.kt
```diff
@@ -155,6 +156,8 @@ private suspend fun OidcConfiguration.executeRequest(
             }.build()
             response = okHttpClient.newCall(newRequest).await()
             if (response.code != rateLimitErrorCode) break
+
+            event = RateLimitExceededEvent(request, response)
```
JayNewstrom, line 160: We need to send this event every time.

That being said, we should probably refactor this to be a do/while instead of dropping into this when we get a 429 (happy to discuss offline).

We also need to add tests to ensure we send the correct amount of events.

auth-foundation/src/main/java/com/okta/authfoundation/client/internal/NetworkUtils.kt
rajdeepnanua-okta: No worries! I missed this also

auth-foundation/src/main/java/com/okta/authfoundation/client/internal/NetworkUtils.kt
rajdeepnanua-okta: I refactored to do/while, and have added tests for number of events as well

auth-foundation/src/main/java/com/okta/authfoundation/client/events/RateLimitExceededEvent.kt
```diff
@@ -44,5 +48,5 @@ class RateLimitExceededEvent internal constructor(
      * Allows the app developer to change the minimum amount of time, in seconds, to wait before
      * retrying [request] when a 429 status code is received.
      */
-    var minDelaySeconds: Long = 5L,
+    var minDelaySeconds: Long = 1L,
```
JayNewstrom, line 51: I think we documented this default somewhere that needs changed.

auth-foundation/src/main/java/com/okta/authfoundation/client/internal/NetworkUtils.kt
```diff
@@ -127,36 +127,38 @@ private suspend fun OidcConfiguration.executeRequest(
...
+        event = RateLimitExceededEvent(requestToExecute, response, retryCount)
         eventCoordinator.sendEvent(event)
+        retryCount++
+
+        val timeToResetSeconds: Long
+        val requestId = response.header("X-Okta-Request-Id")
+        val responseTimeSeconds = response.headers.getDate("Date")?.time?.milliseconds?.inWholeSeconds
+        val retryTimeSeconds = response.header("X-Rate-Limit-Reset")?.toLongOrNull()
+        if (responseTimeSeconds == null || retryTimeSeconds == null) {
+            return response
```
JayNewstrom, line 148: I don't think we want to send the event in this case. Can we postpone the event until after this?

auth-foundation/src/test/java/com/okta/authfoundation/client/NetworkUtilsTest.kt
```diff
@@ -694,61 +694,90 @@ class NetworkingTest {
...
+        val exception = (result as OidcClientResult.Error<String>).exception
+        assertThat(exception).isInstanceOf(OidcClientResult.Error.HttpResponseException::class.java)
+        val httpResponseException = exception as OidcClientResult.Error.HttpResponseException
+        assertThat(httpResponseException.responseCode).isEqualTo(429)
+
+        assertThat(oktaRule.mockWebServerDispatcher.numberRemainingInQueue()).isEqualTo(1)
+        oktaRule.mockWebServerDispatcher.clear()
     }
 
     @Test fun testPerformRequestRetryWithInvalidRateLimitResetHeaderShouldWaitMinDelay(): Unit = runTest {
```
JayNewstrom, line 723: Looks like the name should be updated.

auth-foundation/src/test/java/com/okta/authfoundation/client/NetworkUtilsTest.kt
```diff
@@ -694,61 +694,90 @@ class NetworkingTest {
...
+        val exception = (result as OidcClientResult.Error<String>).exception
+        assertThat(exception).isInstanceOf(OidcClientResult.Error.HttpResponseException::class.java)
+        val httpResponseException = exception as OidcClientResult.Error.HttpResponseException
+        assertThat(httpResponseException.responseCode).isEqualTo(429)
+
+        assertThat(oktaRule.mockWebServerDispatcher.numberRemainingInQueue()).isEqualTo(1)
+        oktaRule.mockWebServerDispatcher.clear()
+    }
+
+    @Test fun testPerformRequestRetrySendsAnEventForEach429Response(): Unit = runTest {
```
JayNewstrom, line 750: We should also make sure the most recent event is the one that's used. So we should change something about it and validate it.

auth-foundation/src/main/java/com/okta/authfoundation/client/events/RateLimitExceededEvent.kt
JayNewstrom: Looks like you caught this!

auth-foundation/src/main/java/com/okta/authfoundation/client/internal/NetworkUtils.kt
rajdeepnanua-okta: Good catch! Yes, I will do that

auth-foundation/src/main/java/com/okta/authfoundation/client/internal/NetworkUtils.kt
```diff
@@ -127,36 +127,40 @@ private suspend fun OidcConfiguration.executeRequest(
 ): Response {
     currentCoroutineContext().ensureActive()
     val rateLimitErrorCode = 429
+    var event: RateLimitExceededEvent
```
JayNewstrom, line 130: Can `event`, and `response` be local `val` inside the `do`?

auth-foundation/src/main/java/com/okta/authfoundation/client/internal/NetworkUtils.kt
```diff
@@ -127,36 +127,40 @@ private suspend fun OidcConfiguration.executeRequest(
...
+        if (response.code != rateLimitErrorCode) return response
+
+        val timeToResetSeconds: Long
+        val requestId = response.header("X-Okta-Request-Id")
+        val responseTimeSeconds = response.headers.getDate("Date")?.time?.milliseconds?.inWholeSeconds
+        val retryTimeSeconds = response.header("X-Rate-Limit-Reset")?.toLongOrNull()
+        if (responseTimeSeconds == null || retryTimeSeconds == null) {
+            return response
+        }
+        response.use {
```
JayNewstrom, line 149: Do we need this `response.use`?

auth-foundation/src/test/java/com/okta/authfoundation/client/NetworkUtilsTest.kt
```diff
@@ -694,61 +695,158 @@ class NetworkingTest {
...
+        assertThat(elapsedTime).isEqualTo((1 + 2 + 4).seconds.inWholeMilliseconds)
+    }
+
+    @Test fun testPerformRequestRetryChangesMaxRetriesBasedOnHandledEvent(): Unit = runTest {
+        val events = mutableListOf<RateLimitExceededEvent>()
+        val rateLimitEventHandler = object : EventHandler {
+            override fun onEvent(event: Any) {
+                val errorEvent = event as RateLimitExceededEvent
+                events.add(event)
+                errorEvent.maxRetries = if (errorEvent.retryCount == 10) 10 else 1000
```
JayNewstrom, line 825: Might be worth adding a comment to this to say what/why we're doing this.

auth-foundation/src/test/java/com/okta/authfoundation/client/NetworkUtilsTest.kt
```diff
@@ -694,61 +695,158 @@ class NetworkingTest {
...
+            }
+        }
+        oktaRule.eventHandler.registerEventHandler(rateLimitEventHandler)
+
+        for (i in 0 until 11) {
```
JayNewstrom, line 830: Also might be worth a comment here explaining why 11. Similar to the above comment.

auth-foundation/src/main/java/com/okta/authfoundation/client/internal/NetworkUtils.kt
rajdeepnanua-okta: There could be a resource leak otherwise. A more explicit way to write what I want is to do ```response.close()``` in a try catch block, if that's preferred

auth-foundation/src/main/java/com/okta/authfoundation/client/internal/NetworkUtils.kt
rajdeepnanua-okta: Response can't be local because I'm returning it outside the while loop. I tried returning it inside the while loop as well, but the compiler thinks it's not good enough to have it there.

auth-foundation/src/main/java/com/okta/authfoundation/client/internal/NetworkUtils.kt
rajdeepnanua-okta: I'm now using a ```response.close()```

auth-foundation/src/test/java/com/okta/authfoundation/client/NetworkUtilsTest.kt
```diff
@@ -694,61 +695,160 @@ class NetworkingTest {
...
+
+        val startTime = currentTime
+        oktaRule.createOidcClient().performRequest(JsonObject.serializer(), request) {
+            it["foo"]!!.jsonPrimitive.content
+        }
         val elapsedTime = currentTime - startTime
 
-        assertThat((result as OidcClientResult.Success<String>).result).isEqualTo("bar")
-        assertThat(elapsedTime).isEqualTo(minDelaySeconds.seconds.inWholeMilliseconds)
+        assertThat(events.size).isEqualTo(4)
```
JayNewstrom, line 815: Let me make sure I'm understanding this right.
We get 4 events. The 3 we delay for (1+2+4 below), and the last one we don't delay for because it would put us over the max retries?

JayNewstrom (approved):

auth-foundation/src/test/java/com/okta/authfoundation/client/NetworkUtilsTest.kt
rajdeepnanua-okta: That's correct

JayNewstrom: 🚢

PR #218, opened 2022-09-15
Add tests for closing responses during rate limit retry

rajdeepnanua-okta:


auth-foundation/src/test/java/com/okta/authfoundation/client/NetworkUtilsTest.kt
```diff
@@ -851,6 +852,51 @@ class NetworkingTest {
...
+            .url(oktaRule.baseUrl.newBuilder().addPathSegments("test").build())
+            .build()
+
+        val response = oktaRule.createOidcClient().performRequest(request)
+
+        var closedResponses = 0
+        assertThat(events.size).isEqualTo(4)
+        val lastRetryEvent = events.removeLast()
+        events.forEach { event ->
+            try {
```
JayNewstrom, line 884: This could be rewritten with `assertFailsWith`, but if not, we'll want to add a `fail` in the catch block as well (if the message isn't closed).

It's not super critical since you're asserting the failed count below. But it's a best practice to fail (which is why the `assertFailsWith` exists!

JayNewstrom (approved):

JayNewstrom (approved):

PR #222, opened 2022-09-20
Fix a potential race when writing exceptionPairs

frett:
a regular MutableMap isn't thread safe, and there was potential for multiple coroutines to update the map concurrently which could corrupt the Map.
This PR utilizes the async task result to generate the Map in a thread-safe manner.


frett:
I didn't run into this issue, but noticed the non-synchronized write while looking at the new `revokeAllTokens()` method.

I think the potential issue would be pretty rare to occur, but could occur on a large enough user base.

JayNewstrom (approved):
Thanks!

PR #240, opened 2023-03-02
Add device token cookie to okHttpClient

rajdeepnanua-okta:
Add DT cookie to http requests. This is mainly useful for okta-idx-android, but there didn't seem to be a clean way to do this by changing okta-idx-android since okHttpClient is already initialized in this SDK before code in idx sdk can be reached.

### DIFF
--- a/auth-foundation/src/main/java/com/okta/authfoundation/client/DeviceTokenCookieJar.kt
+++ b/auth-foundation/src/main/java/com/okta/authfoundation/client/DeviceTokenCookieJar.kt
@@ -18,8 +18,9 @@ package com.okta.authfoundation.client
 import okhttp3.Cookie
 import okhttp3.CookieJar
 import okhttp3.HttpUrl
+import kotlin.time.Duration.Companion.seconds
 
-class DeviceTokenCookieJar : CookieJar {
+class DeviceTokenCookieJar(private val oidcClock: OidcClock) : CookieJar {
     private val savedCookiesCache = mutableMapOf<String, List<Cookie>>()
 
     private val deviceTokenCookieBuilder = Cookie.Builder()
@@ -30,7 +31,7 @@ class DeviceTokenCookieJar : CookieJar {
     override fun loadForRequest(url: HttpUrl): List<Cookie> {
         val deviceTokenCookie = deviceTokenCookieBuilder.domain(url.host).build()
         val savedCookiesForDomain = savedCookiesCache[url.host]?.filter {
-            it.expiresAt > System.currentTimeMillis()
+            it.expiresAt > oidcClock.currentTimeEpochSecond().seconds.inWholeMilliseconds
         } ?: emptyList()
         return savedCookiesForDomain + listOf(deviceTokenCookie)
     }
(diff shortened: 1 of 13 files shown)


auth-foundation/src/main/java/com/okta/authfoundation/client/DeviceTokenCookieJar.kt
```diff
@@ -0,0 +1,45 @@
...
+class DeviceTokenCookieJar : CookieJar {
+    private val deviceTokenCookieBuilder = Cookie.Builder()
+        .name("DT")
+        .value(DeviceTokenProvider.deviceToken)
+        .secure()
+
+    override fun loadForRequest(url: HttpUrl): List<Cookie> {
+        val deviceTokenCookie = deviceTokenCookieBuilder.domain(url.host)
+            .build()
+        return listOf(deviceTokenCookie)
```
mikenachbaur-okta, line 31: Will this _only_ send the device token cookie? Or will it also include cookies previously set by the server? I think the implementation should be that it injects the DT cookie _in addition_ to all regular cookies managed by the device.

auth-foundation/src/main/java/com/okta/authfoundation/client/DeviceTokenCookieJar.kt
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    override fun saveFromResponse(url: HttpUrl, cookies: List<Cookie>) {
+        // Do nothing
+    }
+}
+
+class NoOpCookieJar : CookieJar {
```
mikenachbaur-okta, line 39: Is there a reason for this to be in the SDK itself, if it's only being used from within the tests? For example, could this be put into a shared library the tests can consume?

rajdeepnanua-okta: No particular reason. I just put this here and forgot to move it later.

rajdeepnanua-okta: I've moved it to test-helpers now.

auth-foundation/src/main/java/com/okta/authfoundation/client/DeviceTokenCookieJar.kt
rajdeepnanua-okta: I'm now returning all the cookies saved for the url domain

mikenachbaur-okta:
This looks good, but I'm wondering if wee need to implement something similar to `okta-idx-swift` where a developer can control whether or not the DT cookie is sent. For example, in Swift, a developer could opt out of sending the device token, which is really useful for testing.

Also, do you know whether or not the DT cookie has an impact on any other requests made by our auth pipelines? I'd be concerned that sending this cookie for other requests right cause unexpected side-effects.

rajdeepnanua-okta:
I've tested the changes with browser, resource owner, device authorization, and session token flows. They all seem to be working as expected. All of these flows also have automated E2E tests, except for session token flow.

For configuring whether the DT cookie should be sent, the user can set ```AuthFoundationDefaults.cookieJar = CookieJar.NO_COOKIES``` for the default OkHttp cookie behavior. I can add this detail to the field doc for AuthFoundationDefaults.cookieJar

mikenachbaur-okta (approved):

PR #253, opened 2023-06-13
Update dependencies

rajdeepnanua-okta:


versions.gradle
```diff
@@ -169,7 +178,7 @@ static def addRepos(RepositoryHandler handler) {
 def forceVersions(ConfigurationContainer configurations) {
     configurations.configureEach { configuration ->
         configuration.resolutionStrategy {
-            force deps.bouncy_castle // Vulnerability Fix. OKTA-376968
```
rajdeepnanua-okta, line 172: This is no longer needed because all libraries in the project now depend on a bouncycastle version higher than what's forced. Also, this causes build issues with gradle updates and was painful to find

mikenachbaur-okta (approved):

PR #258, opened 2023-08-02
Decouple ForegroundActivity lifecycle from login state

### LINKED ISSUE #248: "Flow Cancelled" Exception When Activity is Destroyed
### Describe the bug?

When the activity is destroyed due to configuration change, such as device rotation or entering multi-window mode, etc. and the user attempts to login or logout using the chrome custom tab, the user will not be able to complete the operation and the SDK will provide the following exception:

```
Flow cancelled.
                                                                                                    com.okta.webauthenticationui.WebAuthenticationClient$FlowCancelledException: Flow cancelled.
                                                                                                    	at com.okta.webauthenticationui.DefaultRedirectCoordinator.emit(RedirectCoordinator.kt:123)
                                                                                                    	at com.okta.webauthenticationui.SingletonRedirectCoordinator.emit(Unknown Source:2)
                                                                                                    	at com.okta.webauthenticationui.ForegroundViewModel.flowCancelled(ForegroundViewModel.kt:78)
                                                                                                    	at com.okta.webauthenticationui.ForegroundActivity.onPause(ForegroundActivity.kt:67)
                                                                                                    	at android.app.Activity.performPause(Activity.java:8481)
                                                                                                    	at android.app.Instrumentation.callActivityOnPause(Instrumentation.java:1618)
                                                                                                    	at android.app.ActivityThread.performPauseActivityIfNeeded(ActivityThread.java:5054)
                                                                                                    	at android.app.ActivityThread.performPauseActivity(ActivityThread.java:5015)
                                                                                                    	at android.app.ActivityThread.handlePauseActivity(ActivityThread.java:4967)
                                                                                                    	at android.app.servertransaction.PauseActivityItem.execute(PauseActivityItem.java:48)
                                                                                                    	at android.app.servertransaction.ActivityTransactionItem.execute(ActivityTransactionItem.java:45)
                                                                                                    	at android.app.servertransaction.TransactionExecutor.executeLifecycleState(TransactionExecutor.java:176)
                                                                                                    	at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:97)
                                                                                                    	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2307)
                                                                                                    	at android.os.Handler.dispatchMessage(Handler.java:106)
                                                                                                    	at android.os.Looper.loopOnce(Looper.java:201)
                                                                                                    	at android.os.Looper.loop(Looper.java:288)
                                                                                                    	at android.app.ActivityThread.main(ActivityThread.java:7872)
                                                                                                    	at java.lang.reflect.Method.invoke(Native Method)
                                                                                                    	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
                                                                                                    	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:936)
```

### What is expected to happen?

The user is able to login and logout.

### What is the actual behavior?

The SDK throws a `com.okta.webauthenticationui.WebAuthenticationClient$FlowCancelledException`.

### Reproduction Steps?

1. In the Android developer tools, toggle "Don't keep activities" to on. 
2. Attempt to login using the web authentication client

**Actual Results**
The SDK throws a `com.okta.webauthenticationui.WebAuthenticationClient$FlowCancelledException`.

**Expected Results**
The SDK retains state so the user is able to login (perhaps using `onSaveInstanceState`).

### Additional Information?

I suspect this is an issue for multi window devices as well, based on production reports.  Unfortunately I cannot attempt to reproduce.

### SDK Version and Artifact(s) used.

1.12

### Build Information

_No response_

### LINKED ISSUE #257: Make CredentialBootstrap.reset() visible for testing, or make CredentialBootstrap a class instead of object

Writing tests around Okta have been very difficult because of the fact that `CredentialBootstrap` is an `object`, rather than a `class` that can be instantiated. True singletons can be problematic when coding, especially when testing. Because it is an `object`, it doesn't get reset in between tests, but rather sticks around for as long as the program exists in memory, meaning I can't test things that use `CredentialBootstrap` in more than one single test method without them failing. 

I've imported some of the `test-helpers` classes into my own project to perform tests using the Okta system:

```kotlin
class OktaUserAuthenticatorTest {
    @get:Rule
    val okta = OktaRule()

    private val credentialBootstrap by lazy {
        CredentialBootstrap.apply {
            runCatching {
                initialize(
                    okta.createOidcClient(okta.createEndpoints())
                        .createCredentialDataSource(InMemoryTokenStorage())
                )
            }.onFailure { Timber.e("Failed to re-initialize CredentialBootstrap.") }
            runBlocking { defaultCredential().storeToken(null) }
        }
    }
}
```

Note that because of how `CredentialBootstrap` works, I cannot actually call `initialize` more than once without an exception being thrown. The `OidcClient` can only live as long as the `OktaRule`, which only lives as long as one single test method run. Subsequent test methods will get a new rule that can then not be used by `CredentialBootstrap`.

I have now seen that there is `CredentialBootstrap.reset()` that clears the data source, and if that could be made `@VisibleForTesting(otherwise=INTERNAL)`, that would be huge for testing, and should allow me to re-initialize the object with a new data source using the clients created by the `OktaRule` for each test.

`CredentialBootstrap.reset()` should be made visible for testing.

_No response_

### Additional Information?

There are a lot of cool things happening in this library, but I feel like using `objects` is problematic  for a number of reasons, especially dealing with testing. I've been able to get past pretty much all of them except this one.

rajdeepnanua-okta:
This PR fixes #248 to the best of the SDKs abilities. Also this fixes #257, which makes an CredentialBootstrap.reset() visible for user's testing purposes and guards it with InternalAuthFoundationApi annotation.

To fully fix #248, users also need to ensure that they are executing and storing results of call to WebAuthenticationClient.login() in a component that survives activity onDestroy(), rather than storing results in a ViewModel's data holder and executing in viewModelScope.

### DIFF
--- a/web-authentication-ui/src/main/java/com/okta/webauthenticationui/ForegroundViewModel.kt
+++ b/web-authentication-ui/src/main/java/com/okta/webauthenticationui/ForegroundViewModel.kt
@@ -17,48 +17,58 @@ package com.okta.webauthenticationui
 
 import android.app.Activity
 import android.net.Uri
+import android.os.Parcelable
 import androidx.annotation.VisibleForTesting
 import androidx.lifecycle.LiveData
-import androidx.lifecycle.MutableLiveData
+import androidx.lifecycle.SavedStateHandle
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import kotlinx.coroutines.launch
-import okhttp3.HttpUrl
+import kotlinx.parcelize.Parcelize
+import okhttp3.HttpUrl.Companion.toHttpUrl
 
-internal class ForegroundViewModel : ViewModel() {
+internal class ForegroundViewModel(private val savedStateHandle: SavedStateHandle) : ViewModel() {
     companion object {
         // Needs to be inside the companion object due to the tests. Since this is accessed in init, we'd either need to pass it in the
         // constructor, or make it static. Since it's for tests only, making it static was the easier solution, which is why I chose
         // it.
         @VisibleForTesting var redirectCoordinator: RedirectCoordinator = SingletonRedirectCoordinator
+        private const val LIVE_DATA_KEY = "FOREGROUND_STATE_LIVE_DATA_KEY"
     }
 
-    sealed class State {
+    @Parcelize
+    sealed class State : Parcelable {
+        @Parcelize
         object AwaitingInitialization : State()
+        @Parcelize
         object Error : State()
-        class LaunchBrowser(val url: HttpUrl) : State()
+        @Parcelize
+        class LaunchBrowser(val urlString: String) : State()
+        @Parcelize
         object AwaitingBrowserCallback : State()
     }
 
-    private val _stateLiveData = MutableLiveData<State>(State.AwaitingInitialization)
-    val stateLiveData: LiveData<State> = _stateLiveData
+    val stateLiveData: LiveData<State> = savedStateHandle.getLiveData(LIVE_DATA_KEY, State.AwaitingInitialization)
 
     init {
         viewModelScope.launch {
-            when (val result = redirectCoordinator.runInitializationFunction()) {
-                is RedirectInitializationResult.Error -> {
-                    _stateLiveData.value = State.Error
-                }
-                is RedirectInitializationResult.Success -> {
-                    _stateLiveData.value = State.LaunchBrowser(result.url)
+            if (stateLiveData.value == State.AwaitingInitialization) {
+                when (val result = redirectCoordinator.runInitializationFunction()) {
+                    is RedirectInitializationResult.Error -> {
+                        savedStateHandle[LIVE_DATA_KEY] = State.Error
+                    }
+
+                    is RedirectInitializationResult.Success -> {
+                        savedStateHandle[LIVE_DATA_KEY] = State.LaunchBrowser(result.url.toString())
+                    }
                 }
             }
         }
     }
 
-    fun launchBrowser(activity: Activity, url: HttpUrl) {
-        _stateLiveData.value = State.AwaitingBrowserCallback
-        if (!redirectCoordinator.launchWebAuthenticationProvider(activity, url)) {
+    fun launchBrowser(activity: Activity, urlString: String) {
+        savedStateHandle[LIVE_DATA_KEY] = State.AwaitingBrowserCallback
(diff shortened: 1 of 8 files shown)


web-authentication-ui/src/test/java/com/okta/webauthenticationui/ForegroundActivityTest.kt
```diff
@@ -97,26 +108,6 @@ class ForegroundActivityTest {
...
-            on { launchWebAuthenticationProvider(any(), any()) } doReturn true
-            onBlocking { runInitializationFunction() } doAnswer {
-                RedirectInitializationResult.Success(mock(), Any())
-            }
-        }
-        ForegroundViewModel.redirectCoordinator = redirectCoordinator
-        controller.create().resume()
-        assertThat(activity.isFinishing).isFalse()
-        controller.pause().resume()
-        assertThat(activity.isFinishing).isTrue()
```
rajdeepnanua-okta, line 115: Deleted this test case which was mainly testing this flow. ForegroundActivity pausing, then resuming isn't necessarily cancellation of our login flow. ForegroundActivity can be paused while the browser login window is displayed, since it gets backgrounded in that case. Previously we were incorrectly calling finish() in this case, and we are no longer doing that with this PR.

web-authentication-ui/src/main/java/com/okta/webauthenticationui/ForegroundViewModel.kt
```diff
@@ -17,48 +17,58 @@ package com.okta.webauthenticationui
...
+import androidx.lifecycle.SavedStateHandle
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import kotlinx.coroutines.launch
-import okhttp3.HttpUrl
+import kotlinx.parcelize.Parcelize
+import okhttp3.HttpUrl.Companion.toHttpUrl
 
-internal class ForegroundViewModel : ViewModel() {
+internal class ForegroundViewModel(private val savedStateHandle: SavedStateHandle) : ViewModel() {
```
rajdeepnanua-okta, line 30: This is the fix. savedStateHandle survives activity onDestroy() when system kills the activity. This replaces _stateLiveData.

mikenachbaur-okta (approved):

PR #260, opened 2023-08-04
Fix DeviceToken format

rajdeepnanua-okta:
Previously, a randomly generated UUID was being used as device token. This UUID string contained alphanumeric characters, and dashes. Having dashes in device token seems to cause the backend to not remember the device. Removing dashes fixes the issue.


mikenachbaur-okta:
I think the root cause is that the `dt` cookie can have a maximum length of 32 characters, or else the server doesn't understand / remember it. Removing the dashes might have inadvertently brought you under that length.

Can you add a unit test to ensure the character length doesn't exceed 32 characters?

mikenachbaur-okta (approved):

Dark666t (approved):

PR #263, opened 2023-09-21
Add configurable debounce time for login flow cancellation

rajdeepnanua-okta:
This PR adds a debounce time for login flow cancellation.

Some customer devices are encountering an issue where the browser sends multiple empty redirects to us before sending a valid redirect uri during logout. This causes the logout flow to never complete.

To fix this, a configurable debounce time has been added to deal. The default debounce is set to 1 second, which I believe is reasonable since it only affects user-initiated cancellation, or redirect error scenarios.


README.md
```diff
@@ -353,6 +353,13 @@ val oidcConfiguration = OidcConfiguration(
 - java.lang.NoClassDefFoundError: Failed resolution of: Ljava/time/Instant;
   - Fix: configure [Core Library Desugaring](https://developer.android.com/studio/write/java8-support#library-desugaring)
 
+### FlowCancelledException
+
+FlowCancelledException is supposed to be thrown in cases where the user has decided to cancel the login flow, usually by quitting the browser login window. It can sometimes be incorrectly thrown in the following cases:
+- Using Android system webview while logging out. Android webview doesn't store session after a successful login, so logging out never redirects and the user is forced to cancel logout process
+- Deleting browser cache after logging in, and attempting to log out. Similar to the above, it is important for browser to store the login state to logout properly, otherwise browser can not provide logout redirect
+- Browser providing empty redirect results followed by well-defined results. This has been observed in some older devices with older browsers. This can be worked around by setting AuthFoundationDefaults.loginCancellationDebounceTime
```
mikenachbaur-okta, line 361:
```suggestion
FlowCancelledException is supposed to be thrown in cases where the user has decided to cancel the login flow, usually by quitting the browser login window. It can sometimes be incorrectly thrown in the following cases:
- Using the Android system webview while logging out. The webview doesn't store the session after a successful login, so logging out never redirects, and the user is forced to cancel logout process
- Deleting the browser cache after logging in, then attempting to log out. Similar to the above, it is important for browser to store the login state to logout successfully, otherwise the browser can not provide the logout redirect.
- Browser providing empty redirect results, followed by well-defined results. This has been observed in some older devices and browsers. This problem can be worked around by setting AuthFoundationDefaults.loginCancellationDebounceTime
```

auth-foundation/src/main/java/com/okta/authfoundation/AuthFoundationDefaults.kt
```diff
@@ -71,4 +73,9 @@ object AuthFoundationDefaults {
     /** The default [CookieJar]. By default, it adds a DT cookie for identifying the device.
      * To use the default [CookieJar] in OkHttp, set this to [CookieJar.NO_COOKIES] */
     var cookieJar: CookieJar by NoSetAfterGetWithLazyDefaultFactory { DeviceTokenCookieJar(clock) }
+
+    /** The default wait time until web login flow is cancelled after receiving empty redirect response from web browser.
+     * This can resolve some issues related to SDK picking up incorrect redirect results from the web browser, but it adds
+     * delay to browser redirect during errors if set to non-zero value. */
```
mikenachbaur-okta, line 79:
```suggestion
    /** The default wait time until the web login flow is cancelled after receiving empty redirect response from the web browser.
     * This can resolve some issues caused by older devices when invalid redirect results are returned from the older browser. When this is set to a non-zero value, it introduces a
     * delay to all redirects when an error is received. */
```

mikenachbaur-okta (changes requested):
The code looks great, though I'd suggest a few doc updates / wording changes.

mikenachbaur-okta (approved):

PR #266, opened 2023-10-25
Don't retry requests that should be deserialized

rajdeepnanua-okta:


mikenachbaur-okta:
In the Swift SDK we have the [shouldRetry](https://github.com/okta/okta-mobile-swift/blob/e53e2adad5e3faec0d590ab1722afc4728050912/Sources/AuthFoundation/Network/APIClient.swift#L70) delegate function which takes an [enum of available options](https://github.com/okta/okta-mobile-swift/blob/e53e2adad5e3faec0d590ab1722afc4728050912/Sources/AuthFoundation/Network/APIClient.swift#L102). 

Is there a preexisting pattern within the SDK similar to this? I thought we'd added one already, related to max retry counts. This is the mechanism the Swift SDK would use to customize the retry behavior (e.g. detect if it's an error related to SMS rate-limiting, and return a `.doNotRetry` result.

rajdeepnanua-okta:
> In the Swift SDK we have the [shouldRetry](https://github.com/okta/okta-mobile-swift/blob/e53e2adad5e3faec0d590ab1722afc4728050912/Sources/AuthFoundation/Network/APIClient.swift#L70) delegate function which takes an [enum of available options](https://github.com/okta/okta-mobile-swift/blob/e53e2adad5e3faec0d590ab1722afc4728050912/Sources/AuthFoundation/Network/APIClient.swift#L102).
> 
> Is there a preexisting pattern within the SDK similar to this? I thought we'd added one already, related to max retry counts. This is the mechanism the Swift SDK would use to customize the retry behavior (e.g. detect if it's an error related to SMS rate-limiting, and return a `.doNotRetry` result.

We have something similar in the kotlin SDK as well: https://github.com/okta/okta-mobile-kotlin/blob/2431b221c445f78d6ec7ccc36b26d9fa22cc755e/auth-foundation/src/main/java/com/okta/authfoundation/client/events/RateLimitExceededEvent.kt#L4

But, I want to avoid modifying retry logic using that event in the IDX sdk if possible. User-defined event handlers can take precedence over the ones we define. So, our logic can break with user customization in that case. Also, `shouldAttemptJsonDeserialization` is an internal variable only used by IDX, so I think it makes sense to use it for configuring IDX-only behavior in this case.

auth-foundation/src/main/java/com/okta/authfoundation/client/internal/NetworkUtils.kt
```diff
@@ -123,7 +123,8 @@ internal suspend fun <T> OidcConfiguration.internalPerformRequest(
 }
 
 private suspend fun OidcConfiguration.executeRequest(
-    request: Request
+    request: Request,
+    returnVerbatimResponse: (Response) -> Boolean = { false },
```
mikenachbaur-okta, line 127: It feels like `returnVerbatimResponse` doesn't accurately describe what's happening, or that it's explicitly used when a 429 is encountered. I'd suggest calling this igoreRateLimit or shouldRetry, something like that.

auth-foundation/src/main/java/com/okta/authfoundation/client/internal/NetworkUtils.kt
```diff
@@ -106,7 +106,7 @@ internal suspend fun <T> OidcConfiguration.internalPerformRequest(
     currentCoroutineContext().ensureActive()
     return withContext(ioDispatcher) {
         try {
-            val okHttpResponse = executeRequest(request)
+            val okHttpResponse = executeRequest(request, returnVerbatimResponse = shouldAttemptJsonDeserialization)
```
mikenachbaur-okta, line 109: I'm not so sure about the `shouldAttemptJsonDeserialization` name either, but this looks like it's preexisting. Is this the API that's only used by IDX?

It might be nice to be be more explicit about what this is used for.

auth-foundation/src/main/java/com/okta/authfoundation/client/internal/NetworkUtils.kt
rajdeepnanua-okta: I've renamed this to `ignoreRateLimit`, which describes what I'm trying to do a lot better

mikenachbaur-okta (approved):

PR #274, opened 2024-01-15
Define new db for token storage

rajdeepnanua-okta:
This PR defines the database definitions for the new token storage implementation. Currently, it's set to merge into a non-master branch until security provides approval on the design.

### DIFF
--- a/auth-foundation/schemas/com.okta.authfoundation.credential.storage.TokenDatabase/1.json
+++ b/auth-foundation/schemas/com.okta.authfoundation.credential.storage.TokenDatabase/1.json
@@ -1,0 +1,76 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 1,
+    "identityHash": "80e1abffca66c29c27232553ec643603",
+    "entities": [
+      {
+        "tableName": "TokenEntity",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`id` TEXT NOT NULL, `encryptedToken` BLOB NOT NULL, `tags` TEXT NOT NULL, `payloadData` TEXT NOT NULL, `keyAlias` TEXT NOT NULL, `tokenEncryptionType` TEXT NOT NULL, `isDefault` INTEGER NOT NULL, `encryptionExtras` TEXT NOT NULL, PRIMARY KEY(`id`))",
+        "fields": [
+          {
+            "fieldPath": "id",
+            "columnName": "id",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "encryptedToken",
+            "columnName": "encryptedToken",
+            "affinity": "BLOB",
+            "notNull": true
+          },
+          {
+            "fieldPath": "tags",
+            "columnName": "tags",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "payloadData",
+            "columnName": "payloadData",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "keyAlias",
+            "columnName": "keyAlias",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "tokenEncryptionType",
+            "columnName": "tokenEncryptionType",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "isDefault",
+            "columnName": "isDefault",
+            "affinity": "INTEGER",
+            "notNull": true
+          },
+          {
+            "fieldPath": "encryptionExtras",
+            "columnName": "encryptionExtras",
+            "affinity": "TEXT",
+            "notNull": true
+          }
+        ],
+        "primaryKey": {
+          "autoGenerate": false,
+          "columnNames": [
+            "id"
+          ]
+        },
+        "indices": [],
+        "foreignKeys": []
+      }
+    ],
+    "views": [],
+    "setupQueries": [
+      "CREATE TABLE IF NOT EXISTS room_master_table (id INTEGER PRIMARY KEY,identity_hash TEXT)",
+      "INSERT OR REPLACE INTO room_master_table (id,identity_hash) VALUES(42, '80e1abffca66c29c27232553ec643603')"
+    ]
+  }
+}
--- a/auth-foundation/src/main/java/com/okta/authfoundation/credential/storage/TokenEntity.kt
+++ b/auth-foundation/src/main/java/com/okta/authfoundation/credential/storage/TokenEntity.kt
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2024-Present Okta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.okta.authfoundation.credential.storage
+
+import androidx.room.Entity
+import androidx.room.PrimaryKey
+import androidx.room.TypeConverters
+import com.okta.authfoundation.credential.storage.typeconverters.JsonObjectTypeConverter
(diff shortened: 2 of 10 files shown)


auth-foundation/schemas/com.okta.authfoundation.credential.storage.TokenDatabase/1.json
```diff
@@ -0,0 +1,82 @@
...
+          ]
+        },
+        "indices": [],
+        "foreignKeys": []
+      }
+    ],
+    "views": [],
+    "setupQueries": [
+      "CREATE TABLE IF NOT EXISTS room_master_table (id INTEGER PRIMARY KEY,identity_hash TEXT)",
+      "INSERT OR REPLACE INTO room_master_table (id,identity_hash) VALUES(42, '0018cc4d631c8ee5bb0eb7fcb5398040')"
```
mikenachbaur-okta, line 79: What do these values represent?

auth-foundation/schemas/com.okta.authfoundation.credential.storage.TokenDatabase/1.json
```diff
@@ -0,0 +1,82 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 1,
+    "identityHash": "0018cc4d631c8ee5bb0eb7fcb5398040",
+    "entities": [
+      {
+        "tableName": "TokenEntity",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`id` TEXT NOT NULL, `encryptedToken` BLOB NOT NULL, `tags` TEXT NOT NULL, `payloadData` TEXT NOT NULL, `keyAlias` TEXT NOT NULL, `userAuthenticationRequired` INTEGER NOT NULL, `encryptionAlgorithm` TEXT NOT NULL, `isDefault` INTEGER NOT NULL, `encryptionExtras` TEXT NOT NULL, PRIMARY KEY(`id`))",
```
mikenachbaur-okta, line 9: Is "Boolean" not an option for column types?

auth-foundation/schemas/com.okta.authfoundation.credential.storage.TokenDatabase/1.json
```diff
@@ -0,0 +1,82 @@
...
+            "notNull": true
+          },
+          {
+            "fieldPath": "payloadData",
+            "columnName": "payloadData",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "keyAlias",
```
mikenachbaur-okta, line 36: What's the "keyAlias" used for?

auth-foundation/src/main/java/com/okta/authfoundation/credential/storage/TokenEntity.kt
```diff
@@ -0,0 +1,71 @@
...
+    JsonObjectTypeConverter::class
+)
+internal data class TokenEntity(
+    @PrimaryKey
+    val id: String,
+    val encryptedToken: ByteArray,
+    val tags: Map<String, String>,
+    val payloadData: JsonObject,
+    val keyAlias: String,
+    val userAuthenticationRequired: Boolean,
```
mikenachbaur-okta, line 37: What does `userAuthenticationRequired` represent?

auth-foundation/src/main/java/com/okta/authfoundation/credential/storage/TokenEntity.kt
```diff
@@ -0,0 +1,71 @@
...
+    val encryptionAlgorithm: String,
+    val isDefault: Boolean,
```
mikenachbaur-okta, line 39: Is there a circumstance where multiple entries can be marked as `isDefault == true`?

auth-foundation/src/main/java/com/okta/authfoundation/credential/storage/TokenEntity.kt
mikenachbaur-okta, line 37: What do `keyAlias` and `userAuthenticationRequired` do?

auth-foundation/src/main/java/com/okta/authfoundation/credential/storage/TokenEntity.kt
rajdeepnanua-okta: No there isn't. I put isDefault as part of stored db entry for my own convenience. [The followup PR that creates RoomTokenStorage](https://github.com/okta/okta-mobile-kotlin/pull/275) handles the logic for setting, or switching defaults atomically in a single transaction in RoomTokenStorage.replace and RoomTokenStorage.add. I have added unit tests for ensuring this scenario never happens as well.

auth-foundation/src/main/java/com/okta/authfoundation/credential/storage/TokenEntity.kt
rajdeepnanua-okta: Good question. keyAlias is the RSA key that's used for encrypting the token. userAuthenticationRequired means that the user needs to do Biometric or PIN based authentication, in which case we pop up a Biometric prompt.

mikenachbaur-okta: I see.  On Android, or ChromeOS, is there any other kind of user authentication that's possible?  For example, on Apple platforms, you can constrain it to pin, generic biometrics, touchID, faceID, etc.

What I'm getting at is, should this be an enum, or a simple boolean?

rajdeepnanua-okta: Good question. I think I should change this to an enum in case I want to extend the functionality to newer authentication methods in the future. And that also makes decryption a bit more user friendly too.

On the actual Android APIs side, things are a bit messy for Android 10 and below. On Android 11+, the users can specify exactly which combination of methods they want through flags. But, on Android 10 and below, users have to use specific timeouts to choose authentication methods: 0 for Bio + Pin, and -1 for Bio only.

Either way, I think enum should make things a bit simpler to extend in the future. So, I will use that instead.

mikenachbaur-okta: 🤔 With platform version differences, and different capabilities for API revisions, it feels like this part needs to be abstracted away anyway.  Using an enum will certainly help, but perhaps this is an area where we can future-proof the design.

For example, the enum values could be tagged as only supporting certain API versions so the developer gets compiler errors if they use unsupported enum types in their apps.  Or some sort of fallback behavior can be built in to map the behavior for older platforms.

This is similar to the Swift SDK's WebAuthenticationUI handling.  Different iOS versions have different SDKs and classes that handle web-based authentication.  e.g. iOS 12 uses ASWebAuthenticationSession, iOS 11 SFAuthenticationSession, and prior to that plain SFSafariBrowser.  So I created a [WebAuthenticationProvider protocol](https://github.com/okta/okta-mobile-swift/blob/master/Sources/WebAuthenticationUI/Providers/WebAuthenticationProvider.swift) that defined the interface for platform-specific capabilities, and then wrote individual classes that worked on those different platform versions.

That might be overkill for this use-case, but either way, abstracting away the platform- or version-specific handling for something that changes based on API version or environment is useful, not only to deal with the current known conditions, but to future-proof your implementation to support other platforms like ChromeOS or anything else we might nee to support later.

rajdeepnanua-okta: For now, I don't think it's necessary to have platform-specific checks. Android supports either Bio or Pin for authentication. Bio only, or Bio + Pin are supported on all platforms, and I don't see much value in having a Pin only authentication. But, the door for that is open with the enum approach now

auth-foundation/src/main/java/com/okta/authfoundation/credential/storage/TokenEntity.kt
```diff
@@ -0,0 +1,74 @@
...
+    val encryptedToken: ByteArray,
+    val tags: Map<String, String>,
+    val payloadData: JsonObject,
+    val keyAlias: String,
+    val tokenEncryptionType: EncryptionType,
+    val isDefault: Boolean,
+    val encryptionExtras: Map<String, String>
+) {
+    internal enum class EncryptionType {
+        NON_BIO,
```
mikenachbaur-okta, line 42: How about `default` or `standard`? `non_bio` is too specific in my mind, and won't make much sense if other encryption types other than bio and pin are added in the future.

auth-foundation/src/main/java/com/okta/authfoundation/credential/storage/TokenEntity.kt
```diff
@@ -0,0 +1,74 @@
...
+        BIO_ONLY,
+        BIO_AND_PIN
```
mikenachbaur-okta, line 44: Is there a `PIN_ONLY` mode as well?

Issue #276, opened 2024-01-25
Missing capabilities to pass in `state` to OAuth2.0 Authentication Request


haroldteramotosonos:
### Describe the bug?

We need to pass in the `state` value to the AuthorizationCodeFlow to create context from login and authentication result redirect.  However, the `state` value cannot be customized and is always set to a random UUID.

https://github.com/okta/okta-mobile-kotlin/blob/4a2b6e01536feaae5afa11f107e2b9fb24e42cd6/oauth2/src/main/java/com/okta/oauth2/AuthorizationCodeFlow.kt#L114

On Swift okta-mobile-swift, `state` value can be customized and available as part of the constructor.
https://github.com/okta/okta-mobile-swift/blob/5a858e0d700d4c9da5f0a331bf067175d20c566c/Sources/OktaOAuth2/Authentication/AuthorizationCodeFlow.swift#L92

### What is expected to happen?

We would like to have the same capability as okta-mobile-swift and set our own `state` value.

### What is the actual behavior?

`state` value is not customizable and always set to a random UUID.

### Reproduction Steps?

Code sample:
https://github.com/okta/okta-mobile-kotlin/blob/4a2b6e01536feaae5afa11f107e2b9fb24e42cd6/oauth2/src/main/java/com/okta/oauth2/AuthorizationCodeFlow.kt#L114

### SDK Version and Artifact(s) used.

bom "1.1.2"

rajdeepnanua-okta:
Hi @haroldteramotosonos, this seems like a reasonable ask and I will add this to the SDK shortly

rajdeepnanua-okta:
I've merged the changes to allow this. They should be present in 1.2.1

PR #278, opened 2024-01-29
Add state parameter to AuthorizationCodeFlow.start

rajdeepnanua-okta:
This PR adds state parameter to AuthorizationCodeFlow.start as requested by #276.

This PR also fixes some issues related to DeviceTokenCookie initialization with the following:
1. Lazily loading DT cookie in DeviceTokenCookieJar to avoid crashes when users create OidcConfiguration in a Jetpack startup initializer.
2. Attempting to reinitialize DeviceTokenProvider's storage by deleting the storage during Exception, and then recreating the storage.

These should fix most of the issues reported by users. The only exception being if the encryption key is itself corrupted, which can happen and has been reported as an issue for EncryptedSharedPreferences. But, that case should be much rarer than the issues addressed by the above two fixes.

### DIFF
--- a/auth-foundation/src/main/java/com/okta/authfoundation/client/DeviceTokenProvider.kt
+++ b/auth-foundation/src/main/java/com/okta/authfoundation/client/DeviceTokenProvider.kt
@@ -16,12 +16,13 @@
 package com.okta.authfoundation.client
 
 import android.content.Context
+import android.content.SharedPreferences
 import androidx.annotation.VisibleForTesting
 import androidx.security.crypto.EncryptedSharedPreferences
 import androidx.security.crypto.MasterKeys
 import java.util.UUID
 
-class DeviceTokenProvider private constructor(appContext: Context) {
+class DeviceTokenProvider private constructor(private val appContext: Context) {
     internal companion object {
         private const val FILE_NAME = "com.okta.authfoundation.device_token_storage"
         @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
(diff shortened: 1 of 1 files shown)


auth-foundation/src/main/java/com/okta/authfoundation/AuthFoundationDefaults.kt
```diff
@@ -72,7 +71,7 @@ object AuthFoundationDefaults {
 
     /** The default [CookieJar]. By default, it adds a DT cookie for identifying the device.
      * To use the default [CookieJar] in OkHttp, set this to [CookieJar.NO_COOKIES] */
-    var cookieJar: CookieJar by NoSetAfterGetWithLazyDefaultFactory { DeviceTokenCookieJar(clock) }
+    var cookieJar: CookieJar by NoSetAfterGetWithLazyDefaultFactory { CookieJar.NO_COOKIES }
```
mikenachbaur-okta, line 74: How do you propose handling DeviceToken support for IDX? The documentation for this property no longer matches up with the implementation, so I'm wondering what the plan is?

Furthermore, if the cookieJar is set to NO_COOKIES, is this property even needed anymore?

rajdeepnanua-okta: I decided to revert this, and instead fixed the issues with the following:
1. Lazily loading DT cookie in DeviceTokenCookieJar to avoid crashes when users create OidcConfiguration in a Jetpack startup initializer.
2. Attempting to reinitialize DeviceTokenProvider's storage by deleting the storage during Exception, and then recreating the storage.

These should fix most of the issues reported by users. The only exception being if the encryption key is itself corrupted, which can happen and has been reported as an issue for EncryptedSharedPreferences. But, that case should be much rarer than the issues addressed by the above two fixes.

mikenachbaur-okta (approved):

PR #281, opened 2024-02-06
Use new token storage in Credential classes

rajdeepnanua-okta:


auth-foundation-bootstrap/src/main/java/com/okta/authfoundationbootstrap/CredentialBootstrap.kt
```diff
@@ -73,27 +66,27 @@ object CredentialBootstrap {
...
-
-    private suspend fun createDefaultCredential(): Credential {
-        val credential = credentialDataSource.createCredential()
-        credential.storeToken(
-            tags = mapOf(
-                Pair(
-                    CREDENTIAL_NAME_TAG_KEY,
-                    DEFAULT_CREDENTIAL_NAME_TAG_VALUE
-                )
+    suspend fun setDefaultCredential(
```
mikenachbaur-okta, line 73: It feels strange that the token's tags and security needs to be re-assigned in setDefaultCredential as well as storeToken.

Ideally the Credential object should just be assigned to the defaultCredential property, and not have to be re-stored, or have its storage settings changed.  For example, the credential may have already been stored to the encrypted database with some set of security options, but assigning it as the default may cause its security settings to be changed.

The goal of the default credential is to simply mark a credential as being the "current" or default one.  For example:

```
for token in someArrayOfTokens {
  try Credential.store(token, security: [ /* some custom security options */ ])
}

let credentialA = try Credential.find(where: { $0[.sub] == "some-identifier" }).first!
Credential.default = credentialA

/// Some time later ...
let credentialB = try Credential.find(where: { $0[.sub] == "some-other-user-identifier" }).first!
Credential.default = credentialB
```

I don't have to re-store these credentials when I change the default. Instead I can just "toggle" or switch between user credentials.

Furthermore, I can add a notification to detect when the default changes.

```
NotificationCenter.default.addObserver(forName: .defaultCredentialChanged) { notification in
    guard let credential = notification.object as? Credential else { return }

    // Update the UI to show the new user
    self.loggedInAs = credential
}
```

auth-foundation/src/main/java/com/okta/authfoundation/credential/TokenStorage.kt
```diff
@@ -25,33 +27,23 @@ import java.util.Objects
...
-     */
-    suspend fun replace(updatedEntry: Entry)
+    suspend fun replace(
+        id: String,
+        token: Token,
+        metadata: Token.Metadata? = null,
+        security: Credential.Security? = null
+    )
+    suspend fun getToken(id: String, promptInfo: BiometricPrompt.PromptInfo? = Credential.Security.promptInfo): Token
+    suspend fun getDefaultToken(promptInfo: BiometricPrompt.PromptInfo? = Credential.Security.promptInfo): Token?
```
mikenachbaur-okta, line 46: We would need a `setDefaultToken` function as well.

rajdeepnanua-okta:
The requested changes here are provided in https://github.com/okta/okta-mobile-kotlin/pull/292. Relevant sample changes are in https://github.com/okta/okta-mobile-kotlin/pull/293

PR #287, opened 2024-03-15
OIDC configuration refactor

rajdeepnanua-okta:
This PR refactors chunks of OidcConfiguration to be able to store it in a serialized format inside Token

### DIFF
--- a/app/src/main/java/sample/okta/android/SampleApplication.kt
+++ b/app/src/main/java/sample/okta/android/SampleApplication.kt
@@ -43,7 +43,7 @@ class SampleApplication : Application() {
         val oidcConfiguration = OidcConfiguration(
             clientId = BuildConfig.CLIENT_ID,
             defaultScope = SampleHelper.DEFAULT_SCOPE,
-            discoveryUrl = "${BuildConfig.ISSUER}/.well-known/openid-configuration"
+            issuer = BuildConfig.ISSUER
         )
         val oidcClient = OidcClient.createFromConfiguration(oidcConfiguration)
         CredentialBootstrap.initialize(oidcClient.createCredentialDataSource(this))
--- a/app/src/main/java/sample/okta/android/SampleApplication.kt
+++ b/app/src/main/java/sample/okta/android/SampleApplication.kt
@@ -24,7 +24,6 @@ import com.okta.authfoundation.client.OidcConfiguration
 import com.okta.authfoundation.client.SharedPreferencesCache
 import com.okta.authfoundation.credential.CredentialDataSource.Companion.createCredentialDataSource
 import com.okta.authfoundationbootstrap.CredentialBootstrap
-import okhttp3.HttpUrl.Companion.toHttpUrl
 import timber.log.Timber
 
 class SampleApplication : Application() {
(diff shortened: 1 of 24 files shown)


app/src/main/java/sample/okta/android/SampleApplication.kt
```diff
@@ -44,11 +43,9 @@ class SampleApplication : Application() {
         val oidcConfiguration = OidcConfiguration(
             clientId = BuildConfig.CLIENT_ID,
             defaultScope = SampleHelper.DEFAULT_SCOPE,
+            discoveryUrl = "${BuildConfig.ISSUER}/.well-known/openid-configuration"
```
mikenachbaur-okta, line 46: Does it infer the discoveryUrl if it isn't supplied?

auth-foundation/src/main/java/com/okta/authfoundation/client/OidcConfiguration.kt
```diff
@@ -118,10 +128,12 @@ class OidcConfiguration @InternalAuthFoundationApi constructor(
...
 
     /** The Json object to do the decoding from the okta server responses. */
-    @InternalAuthFoundationApi val json: Json = defaultJson()
+    @Transient @InternalAuthFoundationApi val json: Json = defaultJson()
 
-    internal companion object {
+    companion object {
         internal fun defaultJson(): Json = Json { ignoreUnknownKeys = true }
+
+        lateinit var default: OidcConfiguration
```
mikenachbaur-okta, line 136: What is `OidcConfiguration.default` used for / where does its configuration come from?

app/src/main/java/sample/okta/android/SampleApplication.kt
mikenachbaur-okta: [According to the spec](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig), the `/.well-known/openid-configuration` URL path should be appended to the issuer URL. So in the event the discoveryUrl isn't supplied, using the issuer with that path appended to it should be sufficient.

auth-foundation/src/main/java/com/okta/authfoundation/client/OidcConfiguration.kt
mikenachbaur-okta: Is this the equivalent to `okta-mobile-swift`'s Okta.plist shared configuration?

rajdeepnanua-okta: That's correct. It's the same idea as okta-mobile-swift's plist configuration.

app/src/main/java/sample/okta/android/SampleApplication.kt
rajdeepnanua-okta: Should I make discoveryUrl parameter accept either the issuer or the .well-known URL in this case? I think that makes sense

rajdeepnanua-okta: I decided to use issuer instead of discoveryUrl. OidcConfiguration used discoveryUrl previously, but I see that .well-known is always prepended by the issuer. I decided to remove having to add .well-known suffix entirely

PR #288, opened 2024-03-15
App context refactor

rajdeepnanua-okta:
This PR refactors app context to use MutableSharedFlow. This allows classes that depend on Context to wait until the application Context has been captured by ApplicationContextInitializer. This PR also sets the default Cache implementation to SharedPreferencesCache, which is now possible without the users having to explicitly provide Context.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,13 +99,11 @@ import com.okta.authfoundation.AuthFoundationDefaults
 import com.okta.authfoundation.client.OidcClient
 import com.okta.authfoundation.client.OidcClientResult
 import com.okta.authfoundation.client.OidcConfiguration
-import com.okta.authfoundation.client.SharedPreferencesCache
 import com.okta.authfoundation.credential.CredentialDataSource.Companion.createCredentialDataSource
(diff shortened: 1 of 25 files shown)


mikenachbaur-okta (commented):
This LGTM, but it would be nice to get an Android expert to review this further. I like how it makes things (particularly initialization) a bit cleaner.

rajdeepnanua-okta:
After further discussion with wider Android eng group, we decided that it's best to let users provide Context rather than having auto-initialization. There are caveats with auto-initialization which can pose issues:
1. User-defined app initializers can run before our own initializer. If users end up using our APIs before our SDK is initialized, inside their initializer, then that will result in the app crashing.
2. Using a `Flow<Context>` variable instead of `Context` for storing Context avoids the crash. But, it can still result in the app hanging if the user calls `runBlocking { // SDK APIs }` during their initializer. This bug shouldn't happen if the users follow setup properly, but it's a difficult issue to debug, should users run into this. The app never crashes or sends an ANR in this case.

Because of these reasons, it's a better idea to continue having the users explicitly provide Context during SDK initialization.

PR #289, opened 2024-03-15
OIDC client refactor

rajdeepnanua-okta:
This PR removes explicit dependency on users having to interact with oidcClient to create new instances of OAuth flows. Now, the OAuth flows can automatically use OidcClient.default created from OidcConfiguration.default. The followup PR to this will remove CredentialBootstrap altogether, and the only setup step users will need to perform is setting OidcConfiguration.default (and that too will be optional if the user supplies OidcConfiguration to each flow manually).

### DIFF
--- a/app/src/main/java/sample/okta/android/browser/BrowserViewModel.kt
+++ b/app/src/main/java/sample/okta/android/browser/BrowserViewModel.kt
@@ -22,7 +22,7 @@ import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import com.okta.authfoundation.client.OidcClientResult
 import com.okta.authfoundationbootstrap.CredentialBootstrap
-import com.okta.webauthenticationui.WebAuthenticationClient
+import com.okta.webauthenticationui.WebAuthentication
 import kotlinx.coroutines.launch
 import sample.okta.android.BuildConfig
 import sample.okta.android.SampleHelper
@@ -36,14 +36,14 @@ class BrowserViewModel : ViewModel() {
         viewModelScope.launch {
             _state.value = BrowserState.Loading
 
-            val webAuthenticationClient = WebAuthenticationClient()
+            val webAuthentication = WebAuthentication()
             var scope = SampleHelper.DEFAULT_SCOPE
             if (addDeviceSsoScope) {
                 scope += " device_sso"
             }
 
             when (
-                val result = webAuthenticationClient.login(
+                val result = webAuthentication.login(
                     context = context,
                     redirectUrl = BuildConfig.SIGN_IN_REDIRECT_URI,
                     extraRequestParameters = emptyMap(),
--- a/app/src/main/java/sample/okta/android/browser/BrowserViewModel.kt
+++ b/app/src/main/java/sample/okta/android/browser/BrowserViewModel.kt
@@ -22,7 +22,7 @@ import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import com.okta.authfoundation.client.OidcClientResult
 import com.okta.authfoundationbootstrap.CredentialBootstrap
-import com.okta.webauthenticationui.WebAuthenticationClient.Companion.createWebAuthenticationClient
+import com.okta.webauthenticationui.WebAuthenticationClient
 import kotlinx.coroutines.launch
 import sample.okta.android.BuildConfig
 import sample.okta.android.SampleHelper
@@ -36,7 +36,7 @@ class BrowserViewModel : ViewModel() {
         viewModelScope.launch {
             _state.value = BrowserState.Loading
 
-            val webAuthenticationClient = CredentialBootstrap.oidcClient.createWebAuthenticationClient()
+            val webAuthenticationClient = WebAuthenticationClient()
             var scope = SampleHelper.DEFAULT_SCOPE
             if (addDeviceSsoScope) {
                 scope += " device_sso"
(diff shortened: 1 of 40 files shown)


app/src/main/java/sample/okta/android/browser/BrowserViewModel.kt
```diff
@@ -36,7 +36,7 @@ class BrowserViewModel : ViewModel() {
         viewModelScope.launch {
             _state.value = BrowserState.Loading
 
-            val webAuthenticationClient = CredentialBootstrap.oidcClient.createWebAuthenticationClient()
+            val webAuthenticationClient = WebAuthenticationClient()
```
mikenachbaur-okta, line 39: Did we want to consider making the class name similar to the Swift [WebAuthentication](https://okta.github.io/okta-mobile-swift/development/webauthenticationui/documentation/webauthenticationui/webauthentication)?

Alternatively, we might want to discuss how these UI environment abstraction classes will function in the future, particularly since we're going to be adding multiple for the Javascript client SDK (e.g. Vue, React, React Native, Electron, etc)

app/src/main/java/sample/okta/android/dashboard/DashboardViewModel.kt
```diff
@@ -119,7 +119,7 @@ internal class DashboardViewModel(private val credentialTagNameValue: String?) :
         viewModelScope.launch {
             val idToken = credential.token?.idToken ?: return@launch
             when (
-                val result = CredentialBootstrap.oidcClient.createWebAuthenticationClient().logoutOfBrowser(
+                val result = WebAuthenticationClient().logoutOfBrowser(
```
mikenachbaur-okta, line 122: Is the recommended pattern to create a client instance every time? Or will there be a shared singleton instance of the default WebAuthentication class like Swift's [WebAuthentication.shared](https://okta.github.io/okta-mobile-swift/development/webauthenticationui/documentation/webauthenticationui/webauthentication/shared) static property. 

The Swift SDK did it this way so listeners/delegates can be added to the shared instance, and would persist across interactions with it.

auth-foundation/src/main/java/com/okta/authfoundation/client/OidcConfiguration.kt
```diff
@@ -132,8 +132,18 @@ class OidcConfiguration private constructor(
...
+        /**
+         * The default OidcConfiguration. This must be set before calling any OAuth flows. Note that this variable
+         * can only be set once in the lifetime of the app.
+         */
+        var default: OidcConfiguration
+            get() = _default ?: throw IllegalStateException("Attempted to use OidcConfiguration.default without setting it")
+            set(value) {
+                if (_default == null) _default = value
+                else throw IllegalStateException("Attempted setting OidcConfiguration.default after initialization")
+            }
```
mikenachbaur-okta, line 146: Are we planning on building out a configuration file, similar to the Swift version, that can implicitly load the default configuration?

web-authentication-ui/src/main/java/com/okta/webauthenticationui/WebAuthenticationClient.kt
```diff
@@ -40,31 +38,40 @@ import com.okta.webauthenticationui.events.CustomizeCustomTabsEvent
...
+     * @param webAuthenticationProvider the [WebAuthenticationProvider] which will be used to show the UI when performing the
+     * redirect flows.
+     */
+    constructor(
+        oidcConfiguration: OidcConfiguration,
+        webAuthenticationProvider: WebAuthenticationProvider = DefaultWebAuthenticationProvider(oidcConfiguration.eventCoordinator)
+    ) : this(OidcClient.createFromConfiguration(oidcConfiguration), webAuthenticationProvider)
+
+    private val authorizationCodeFlow: AuthorizationCodeFlow = AuthorizationCodeFlow(oidcClient)
+    private val redirectEndSessionFlow: RedirectEndSessionFlow = RedirectEndSessionFlow(oidcClient)
```
mikenachbaur-okta, line 74: You might want to make the getter for these public, so a developer can hook into the event lifecycle / customize the behavior of the underlying flows.  This allows a developer to use the higher-level WebAuthentication class, while still being able to tweak / monitor the underlying details, without having to rebuild everything from scratch.

mikenachbaur-okta (commented):
I really like these updates, it definitely makes it feel more developer-friendly!

app/src/main/java/sample/okta/android/browser/BrowserViewModel.kt
rajdeepnanua-okta: Sure, I can rename it to WebAuthentication for consistency.

app/src/main/java/sample/okta/android/dashboard/DashboardViewModel.kt
rajdeepnanua-okta: There is one (internally set) listener that the WebAuthentication class references, but that itself is Singleton. Users get the result of WebAuthentication, while we handle the listeners internally. Since listeners can't be set by the users for WebAuthentication, I don't see much value in adding a shared static property here.

auth-foundation/src/main/java/com/okta/authfoundation/client/OidcConfiguration.kt
rajdeepnanua-okta: I looked into this a bit, but I think I should handle this in a separate PR. There are a few different ways of achieving this, and I need to dig into this more to see what would be ideal. All the options I found so far have some downsides. Adding config file wouldn't be a breaking API change.

PR #292, opened 2024-04-08
Move user-facing credential APIs to Credential

rajdeepnanua-okta:
This PR moves the user-facing credential APIs from CredentialDataSource and CredentialBootstrap to Credential. This ends up removing the authfoundation-bootstrap module entirely, and makes CredentialDataSource an internal class.


auth-foundation/src/main/java/com/okta/authfoundation/client/EncryptionTokenProvider.kt
```diff
@@ -21,49 +21,33 @@ import androidx.datastore.preferences.core.Preferences
...
 import androidx.datastore.preferences.core.stringPreferencesKey
 import androidx.datastore.preferences.preferencesDataStore
-import com.okta.authfoundation.InternalAuthFoundationApi
 import com.okta.authfoundation.util.AesEncryptionHandler
 import kotlinx.coroutines.flow.firstOrNull
 import java.util.UUID
 
-@InternalAuthFoundationApi
-class DeviceTokenProvider(
+internal class EncryptionTokenProvider(
```
rajdeepnanua-okta, line 28: I decided to use a completely different token for DB encryption compared to the one used for DT cookie. It seemed like an unnecessary risk.

auth-foundation/src/main/java/com/okta/authfoundation/client/OidcConfiguration.kt
```diff
@@ -74,10 +73,6 @@ class OidcConfiguration private constructor(
     /** The factory for creating a new instance of [Cache]. [Cache] is used to optimize network calls by the SDK. */
     @property:InternalAuthFoundationApi
     val cacheFactory: suspend () -> Cache = AuthFoundationDefaults.cacheFactory,
-
-    /** The CookieJar used for the network calls used by the SDK. */
-    @property:InternalAuthFoundationApi
-    val cookieJar: CookieJar = AuthFoundationDefaults.cookieJar
```
rajdeepnanua-okta, line 80: Removing this property access here allows us to set cookieJar after OidcConfiguration has been created (and before any network calls are made). This is useful mostly for IDX

auth-foundation/src/main/java/com/okta/authfoundation/credential/Credential.kt
```diff
@@ -137,14 +215,29 @@ class Credential internal constructor(
...
      */
     val tags: Map<String, String>
         get() {
             return Collections.unmodifiableMap(state.value.tags)
         }
 
+    /**
+     * Set tags for this [Credential].
+     */
+    suspend fun setTags(tags: Map<String, String>) {
```
rajdeepnanua-okta, line 227: Before, this functionality was provided by storeToken function. storeToken used to be replaceToken + setTags. I decided to separate them out.

auth-foundation/src/main/java/com/okta/authfoundation/credential/Credential.kt
```diff
@@ -490,7 +527,7 @@ private sealed class CredentialState {
         override val tags: Map<String, String>
     ) : CredentialState()
 
-    object Deleted : CredentialState()
+    data object Deleted : CredentialState()
```
rajdeepnanua-okta, line 530: Changing object to data object doesn't do anything meaningful for our use case, but it's good practice to use it for sealed classes (and for suppressing IDE lint in this case)

auth-foundation/src/main/java/com/okta/authfoundation/credential/CredentialDataSource.kt
```diff
@@ -15,187 +15,85 @@
...
-        val payloadData = idToken?.deserializeClaims(JsonObject.serializer())
-        val previousMetadata = metadata(id) ?: throw IllegalStateException("Token metadata for $id not found")
-        val newMetadata = previousMetadata.copy(
-            tags = tags,
-            payloadData = payloadData,
-            isDefault = isDefault ?: previousMetadata.isDefault
-        )
-        storage.replace(id, token, newMetadata, security)
-    }
+    suspend fun replaceToken(id: String, token: Token) = storage.replace(id, token)
```
rajdeepnanua-okta, line 81: A lot of the logic in replace earlier was for handling if the user called CredentialDataSource.replace directly, or if it was called from a Credential object. This is all that's needed now since it's always called internally.

auth-foundation/src/main/java/com/okta/authfoundation/credential/Credential.kt
```diff
@@ -106,6 +105,85 @@ class Credential internal constructor(
...
+        private val defaultCredentialIdDataStore: DefaultCredentialIdDataStore
+            get() = DefaultCredentialIdDataStore.instance
+
+        @VisibleForTesting
+        internal suspend fun credentialDataSource() = CredentialDataSource.getInstance()
+
+        /**
+         * Returns the default [Credential], or null if no default [Credential] exists.
+         */
+        suspend fun getDefaultCredential(): Credential? {
```
mikenachbaur-okta, line 118: Is it common practice in Kotlin/Java to have different getter/setter functions? Or is it common practice to create a var with custom getter/setter functions?

I'm wondering because it feels like it would be more convenient to be able to say `Credential.default` instead of `Credential.getDefaultCredential()`

auth-foundation/src/main/java/com/okta/authfoundation/credential/Credential.kt
```diff
@@ -106,6 +105,85 @@ class Credential internal constructor(
...
+         */
+        suspend fun setDefaultCredential(credential: Credential) {
+            defaultCredentialIdDataStore.setDefaultCredentialId(credential.id)
+            AuthFoundationDefaults.eventCoordinator.sendEvent(DefaultCredentialChangedEvent(credential))
+        }
+
+        /**
+         * Returns IDs of all available [Credential] objects in storage.
+         */
+        suspend fun allIds(): List<String> = credentialDataSource().allIds()
```
mikenachbaur-okta, line 133: Likewise, instead of a function, is it possible for this to be a `var` with a custom getter?

auth-foundation/src/main/java/com/okta/authfoundation/credential/Credential.kt
```diff
@@ -137,14 +215,29 @@ class Credential internal constructor(
...
     val tags: Map<String, String>
         get() {
             return Collections.unmodifiableMap(state.value.tags)
         }
 
+    /**
+     * Set tags for this [Credential].
+     */
+    suspend fun setTags(tags: Map<String, String>) {
+        token?.let {
```
mikenachbaur-okta, line 228: Why is token optional here? Is it possible for a Credential's token value to be null?

auth-foundation/src/main/java/com/okta/authfoundation/credential/Credential.kt
```diff
@@ -106,6 +105,85 @@ class Credential internal constructor(
         }
     }
 
+    companion object {
+        private val defaultCredentialIdDataStore: DefaultCredentialIdDataStore
+            get() = DefaultCredentialIdDataStore.instance
+
+        @VisibleForTesting
+        internal suspend fun credentialDataSource() = CredentialDataSource.getInstance()
```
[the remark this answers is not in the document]
rajdeepnanua-okta: I initially tried to make this a lazy assigned property. The problem is that lazy properties can't access coroutine methods. TokenStorage is constructed in a suspend function, so it accesses coroutine functionality. In CredentialDataSource.getInstance(), I use a mutex to ensure that only one instance can be constructed.

auth-foundation/src/main/java/com/okta/authfoundation/credential/Credential.kt
rajdeepnanua-okta: This stems from a limitation. Default credential is fetched using coroutine APIs, but properties can't access coroutine APIs without using runBlocking.

auth-foundation/src/main/java/com/okta/authfoundation/credential/Credential.kt
rajdeepnanua-okta: Credential's token value can be null after it has been deleted. In that case, if the user tries setTags, this method throws. I just checked the swift SDK, and I see that it doesn't delete the token stored in the Credential object, but instead removes it from storage and CredentialDataSource. Should I do something similar here? That makes things simpler to maintain on my end.

auth-foundation/src/main/java/com/okta/authfoundation/client/OidcClient.kt
```diff
@@ -279,7 +279,7 @@ class OidcClient private constructor(
                 try {
                     TokenValidator(this@OidcClient, token, nonce, maxAge, jwksDeferred.await()).validate()
                     configuration.eventCoordinator.sendEvent(TokenCreatedEvent(token, credential))
-                    credential?.storeToken(token)
+                    credential?.replaceToken(token)
```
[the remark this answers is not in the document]
rajdeepnanua-okta: Yeah, OidcClient has an instance to Credential in this case. I will propagate this change to Credential to instead fetch the value from OAuth2Client, and update its own value.

mikenachbaur-okta (approved):
This looks good to me!

PR #294, opened 2024-04-10
Add automatic token storage migration

rajdeepnanua-okta:


mikenachbaur-okta (approved):
Does it make sense to move the V1StoreMigrator into a separate namespace? If we create a new migrator in the future, or if we don't want code auto-completion to be cluttered with other types, that might be useful.

Overall it looks great, I just want to be aware of naming.

PR #298, opened 2024-05-09
Add biometrics with timeout

rajdeepnanua-okta:
This PR adds Biometrics with timeout, and sets that as default. This makes things convenient for users when performing multiple get operations on Biometric Credentials. Auth-per-use Biometric Credentials can still be used by users if they explicitly set timeout to 0.


mikenachbaur-okta (approved):
LGTM though I'd like to see what others more familiar with Android have to say

FeiChen-okta (approved):

PR #299, opened 2024-05-10
Add sync Credential APIs, and suffix async ones with Async

rajdeepnanua-okta:


README.md
```diff
@@ -232,15 +232,15 @@ There are multiple terms that might be confused when logging a user out.
 ### Using a Credential to determine user authentication status
 There are a few options to determine the status of a user authentication. Each option has unique pros and cons and should be chosen based on the needs of your use case.
 
-- Non null default credential: `Credential.getDefaultCredential() != null`
+- Non null default credential: `Credential.getDefaultAsync() != null`
```
mikenachbaur-okta, line 235: I think, for consistency, we should reference the sync versions, ala `Credential.default`, etc.  It makes the code more readable, and during the natural course of the SDK, the async version is unnecessary (e.g. it doesn't go to storage with every single call, since the credential instance is cached in-memory)

README.md
```diff
@@ -257,6 +257,7 @@ The SDK has built-in support for handling Biometric encryption. To set the defau
 > Notes:
 > - The SDK does not check which biometrics are enrolled on the user's device. Please check this using https://developer.android.com/reference/android/hardware/biometrics/BiometricManager#canAuthenticate(int) before setting the appropriate security level
 > - The SDK automatically deletes Token entries stored using invalidated biometric keys.
+> - Biometric Credentials should only be fetched using async APIs in `Credential` class.
```
mikenachbaur-okta, line 260: Does the sync version throw an exception when biometrics are required? If so, you might want to indicate that, or reference the specific exception that is thrown, or maybe reference the API docs where this detail can be included.

README.md
```diff
@@ -403,10 +404,12 @@ V1ToV2StorageMigrator.legacyStorage = legacyTokenStorage
...
+1.x would create a new `Credential` with null `Token` if no default `Credential` existed when calling `CredentialBootstrap.defaultCredential()`. In 2.x, default `Credential` can be fetched using `Credential.default` or `Credential.getDefaultAsync()`. Both of those have a type of `Credential?` instead of `Credential`, and return `null` if no default `Credential` exists.
 
 1.x contained `Credential` handling APIs in `CredentialBootstrap` and `CredentialDataSource`. All `Credential` management calls have been moved to `Credential` in 2.x. `CredentialBootstrap` has been deleted, and `CredentialDataSource` is private in 2.x.
 
+1.x provided `suspend` functions for handling creation and management of any `Credential`s. 2.x provides synchronous `Credential` management functions in addition to `suspend` functions. Note that it's still recommended to use `suspend` functions (suffixed with `Async`) if a coroutine scope is available.
```
mikenachbaur-okta, line 411: Is there a particular reason why using the suspend functions are preferred?

mikenachbaur-okta (approved):
LGTM. I'm wondering what the performance implications would be of using the sync variants of these functions, beyond the initial fetch from the database. 

The vast majority of interactions with a Credential will be once the object has already been fetched from the database, meaning the object will have already been created and will reside in memory.

README.md
rajdeepnanua-okta: Good point. I will change this to Credential.default

README.md
rajdeepnanua-okta: It does. I will make this change to specify the exception thrown.

README.md
rajdeepnanua-okta: I wrote this defensively to prevent crashes when using biometrics. But, I think the exception gets the point across, so I will remove the note

mikenachbaur-okta (approved):

PR #313, opened 2024-09-12
Fix corrupt encrypted files caused by unwanted backup and restore

rajdeepnanua-okta:
This PR adds recovery mechanisms for handling unexpected backup and restore with SDK storage files. Since SDK files in storage are encrypted using device keystore, they shouldn't be copied over to other devices, or even across app installations.

In addition to the recovery mechanisms, this PR also adds rules to exclude SDK files from automatic backup. I tested backup and restore using this to ensure sdk files aren't getting backed up: https://developer.android.com/identity/data/testingbackup

### DIFF
--- a/auth-foundation/src/main/java/com/okta/authfoundation/credential/TokenDbRecoveryUtil.kt
+++ b/auth-foundation/src/main/java/com/okta/authfoundation/credential/TokenDbRecoveryUtil.kt
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2024-Present Okta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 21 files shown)


auth-foundation/src/main/java/com/okta/authfoundation/credential/TokenDbRecoveryUtil.kt
```diff
@@ -0,0 +1,34 @@
...
+        val defaultUncaughtExceptionHandler = Thread.getDefaultUncaughtExceptionHandler()
+        Thread.setDefaultUncaughtExceptionHandler { thread, exception ->
+            if (exception is SQLiteException) {
+                context.deleteDatabase(TokenDatabase.DB_NAME)
+            } else {
+                defaultUncaughtExceptionHandler?.uncaughtException(thread, exception)
+            }
+        }
+    }
+}
```
rajdeepnanua-okta, line 34: I decided to add this just in case things don't get fixed with the other fixes. This code should fix the token database if it ends up in an unrecoverable state, even if the encryption token is unchanged.

FeiChen-okta (approved):

PR #314, opened 2024-09-12
Initialize token database before default id store

rajdeepnanua-okta:
This PR adds a fix for the scenario where user migrates from v1 of this SDK to v2. For users migrating from v1 to v2, automigration for storage kicks off after Token storage is initialized. But, DefaultCredentialIdDataStore could be accessed before token storage was migrated. Because of this. Credential.default would always return null on the first call after v1 -> v2 migration.

To fix this, this PR simply calls AuthfoundationDefaults.tokenStorageFactory in DefaultCredentialIdDataStore when getting the default credential. The contract for AuthfoundationDefaults.tokenStorageFactory says that it should return the singleton token storage, so calling it repeatedly for each call to getDefaultCredentialId should have no negative impact.

### DIFF
--- a/auth-foundation/src/main/java/com/okta/authfoundation/AuthFoundation.kt
+++ b/auth-foundation/src/main/java/com/okta/authfoundation/AuthFoundation.kt
@@ -22,4 +22,8 @@ object AuthFoundation {
     fun initializeAndroidContext(context: Context) {
         ApplicationContextHolder.setApplicationContext(context)
     }
+
+    internal suspend fun initializeStorage() {
+        AuthFoundationDefaults.tokenStorageFactory()
+    }
 }
--- a/auth-foundation/src/main/java/com/okta/authfoundation/credential/DefaultCredentialIdDataStore.kt
+++ b/auth-foundation/src/main/java/com/okta/authfoundation/credential/DefaultCredentialIdDataStore.kt
@@ -21,6 +21,7 @@ import androidx.datastore.preferences.core.Preferences
 import androidx.datastore.preferences.core.edit
 import androidx.datastore.preferences.core.stringPreferencesKey
 import androidx.datastore.preferences.preferencesDataStore
+import com.okta.authfoundation.AuthFoundation
 import com.okta.authfoundation.client.ApplicationContextHolder
 import com.okta.authfoundation.util.AesEncryptionHandler
 import kotlinx.coroutines.flow.firstOrNull
@@ -38,6 +39,7 @@ internal class DefaultCredentialIdDataStore(
     private val context by lazy { ApplicationContextHolder.appContext }
 
     suspend fun getDefaultCredentialId(): String? {
+        AuthFoundation.initializeStorage()
         val encryptedDefaultCredentialId = context.dataStore.data.firstOrNull()?.get(PREFERENCE_KEY)
         return encryptedDefaultCredentialId?.let {
             aesEncryptionHandler.decryptString(it).getOrNull()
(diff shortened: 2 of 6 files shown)


mikenachbaur-okta:
So the Kotlin SDK's token storage performs an automatic migration of all stored credentials at app launch? Or does it migrate tokens on-demand as they're accessed?

rajdeepnanua-okta:
> So the Kotlin SDK's token storage performs an automatic migration of all stored credentials at app launch? Or does it migrate tokens on-demand as they're accessed?

It does an automatic migration of all stored credentials

auth-foundation/src/main/java/com/okta/authfoundation/AuthFoundation.kt
```diff
@@ -22,4 +22,8 @@ object AuthFoundation {
     fun initializeAndroidContext(context: Context) {
         ApplicationContextHolder.setApplicationContext(context)
     }
+
+    internal suspend fun initializeStorage() {
+        AuthFoundationDefaults.tokenStorageFactory()
```
mikenachbaur-okta, line 27: So you're touching the factory and ignoring its response object to ensure that it is initialized?

auth-foundation/src/main/java/com/okta/authfoundation/credential/DefaultCredentialIdDataStore.kt
```diff
@@ -38,6 +39,7 @@ internal class DefaultCredentialIdDataStore(
     private val context by lazy { ApplicationContextHolder.appContext }
 
     suspend fun getDefaultCredentialId(): String? {
+        AuthFoundation.initializeStorage()
```
mikenachbaur-okta, line 42: Is the default credential ID not stored within the token database?  Or is it stored elsewhere?

rajdeepnanua-okta: It's stored in this DataStore separate from the TokenDatabase

auth-foundation/src/main/java/com/okta/authfoundation/AuthFoundation.kt
rajdeepnanua-okta: Yeah exactly

mikenachbaur-okta (dismissed):
I think this is good for now, but I think having two separate datastores to manage dependent data feels wrong, unless there is some closer relationship established between the two.

The TokenStorage interface is meant to provide thread-safe interactions with the lower-level token store, encapsulating the relationship between the IDs, tokens, and metadata.  If those values are stored separately that's fine, but I feel like the stability of that thread-safe interface should be preserved.

I think we should file a ticket to revisit this in the future, such that setting, changing, or removing a default token is properly synchronized across the multiple files it uses, and to ensure a consistent interface. If this were the case, then the default ID store wouldn't need to be initialized separately from the database.

rajdeepnanua-okta:
> I think this is good for now, but I think having two separate datastores to manage dependent data feels wrong, unless there is some closer relationship established between the two.
> 
> The TokenStorage interface is meant to provide thread-safe interactions with the lower-level token store, encapsulating the relationship between the IDs, tokens, and metadata. If those values are stored separately that's fine, but I feel like the stability of that thread-safe interface should be preserved.
> 
> I think we should file a ticket to revisit this in the future, such that setting, changing, or removing a default token is properly synchronized across the multiple files it uses, and to ensure a consistent interface. If this were the case, then the default ID store wouldn't need to be initialized separately from the database.

I agree. I think the storing default token in a separate storage or table makes sense since default token ID is a separate entry from the stored tokens in TokenDatabase. But, it might be better to define default token store to be a a different table in Token database rather than its own data store. I'll revisit this in the future

mikenachbaur-okta:
Yeah, and since you're in a SQL database, you could leverage transactions to help avoid some threading issues...

PR #315, opened 2024-09-12
Allow using accessToken if idToken is missing

rajdeepnanua-okta:


mikenachbaur-okta:
It might be worthwhile to follow a more explicit approach, like we use in the Swift (and now Javascript) SDKs.  Basically, when the token is deserialized from the server-parsed JSON, we use the `Date.nowCoordinated` value and save it as the `issuedAt` date. https://github.com/okta/okta-mobile-swift/blob/3ca3e3b549d587c19c16349d9ac02a6871baa85a/Sources/AuthFoundation/Token%20Management/Token%2BInitialization.swift#L18

We could always support migrations from older SDK versions by falling back to the ID token value, but we should still store/encode the actual date in the token storage so we can have an explicit value to use.

rajdeepnanua-okta:
> It might be worthwhile to follow a more explicit approach, like we use in the Swift (and now Javascript) SDKs. Basically, when the token is deserialized from the server-parsed JSON, we use the `Date.nowCoordinated` value and save it as the `issuedAt` date. https://github.com/okta/okta-mobile-swift/blob/3ca3e3b549d587c19c16349d9ac02a6871baa85a/Sources/AuthFoundation/Token%20Management/Token%2BInitialization.swift#L18
> 
> We could always support migrations from older SDK versions by falling back to the ID token value, but we should still store/encode the actual date in the token storage so we can have an explicit value to use.

I've made changes to make the logic simpler. Token now has a public constructor which doesn't take in an issuedAt parameter, and that's used during client deserialization. If the Token object in storage doesn't have an issuedAt, it gets initialized with an expired issuedAt time through the private constructor during deserialization from storage.

So this way any callers that create a Token object are guaranteed to have the correct issuedAt time.

auth-foundation/src/main/java/com/okta/authfoundation/credential/Token.kt
```diff
@@ -66,7 +69,34 @@ class Token constructor(
      * The configuration that was used to fetch this token.
      */
     val oidcConfiguration: OidcConfiguration,
+    @EncodeDefault
+    internal val issuedAt: Long = oidcConfiguration.clock.currentTimeEpochSecond() - expiresIn,
```
mikenachbaur-okta, line 73: And to clarify, this value is only used when `issuedAt` is not supplied to the constructor?  Or is this only used in serialization decoding?

auth-foundation/src/main/java/com/okta/authfoundation/credential/Token.kt
```diff
@@ -187,3 +234,8 @@ internal class SerializableToken internal constructor(
         )
     }
 }
+
+@Serializable
+private class TokenIssuedAtPayload(
```
mikenachbaur-okta, line 239: I'd suggest renaming this to something to signify why it's being used.  In my mind it feels like this, and the `getIssuedTime` function, are there to extract values from the server response necessary for storage or validation of the token.  So `TokenInitializationPayload` or something?

auth-foundation/src/main/java/com/okta/authfoundation/credential/Token.kt
rajdeepnanua-okta: This value is only used in serialization decoding. The other constructor is the one that's publicly supplied to the rest of the SDK

auth-foundation/src/main/java/com/okta/authfoundation/credential/Token.kt
rajdeepnanua-okta: I've renamed it to TokenInitializationPayload

mikenachbaur-okta (approved):
Looks good! :D

PR #323, opened 2025-01-30
Make AuthFoundationDefaults configurable by other SDKs

rajdeepnanua-okta:
This PR does the following:
1. Makes AuthFoundationDefaults configurable by other SDKs (without the risk of crashing the app)
2. Update libraries to the latest versions available. The only exception being spotless and jackson, which runs into issues with newer versions
3. Add snyk scan. This is mostly the same implementation as in [okta-devices-kotlin](https://github.com/okta/okta-devices-kotlin).


scripts/dependency_scan.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+#####
+## Build Environment Setup
+#####
+java17_0
```
mikenachbaur-okta, line 6: What's this do?

mikenachbaur-okta (approved):

scripts/dependency_scan.sh
rajdeepnanua-okta: This sets up java 17 in our internal build system

jaredperreault-okta (approved):

PR #345, opened 2025-09-18
Initial direct-auth project structure

FeiChen-okta:
Add initial interfaces to auth foundation and okta-direct-auth
RESOLVES: OKTA-1008730


auth-foundation/src/main/java/com/okta/authfoundation/GrantType.kt
```diff
@@ -0,0 +1,125 @@
...
+     * @see [Okta Direct Authentication WebAuthn](https://developer.okta.com/docs/reference/direct-auth/grant-types/#webauthn)
+     */
+    object WebAuthn : GrantType("urn:okta:params:oauth:grant-type:webauthn")
+
+    /**
+     * A grant type for Multi-Factor Authentication (MFA) using WebAuthn.
+     *
+     * @see [Okta MFA WebAuthn](https://developer.okta.com/docs/guides/configure-direct-auth-grants/bmfawebauthn/main/#about-the-direct-authentication-grant)
+     */
+    object WebAuthnMfa : GrantType("urn:okta:params:oauth:grant-type:mfa-webauthn")
```
rajdeepnanua-okta: I see a lot of repeated MFA/non-MFA types. Should we consider MFA a field in data class for such cases?

auth-foundation/src/main/java/com/okta/authfoundation/api/http/log/AuthFoundationLogger.kt
```diff
@@ -0,0 +1,37 @@
...
+ */
+interface AuthFoundationLogger {
+    /**
+     * Writes a log message with an optional throwable and a specified log level.
+     *
+     * @param message The log message to be printed.
+     * @param tr An optional [Throwable] to be logged with the message. Defaults to null.
+     * @param logLevel The severity of the log message. Defaults to [LogLevel.INFO].
+     */
+    fun write(
```
rajdeepnanua-okta: Why not call this log?

oauth2/src/test/java/com/okta/oauth2/DeviceAuthorizationFlowTest.kt
```diff
@@ -63,7 +66,7 @@ class DeviceAuthorizationFlowTest {
                 method("POST"),
                 path("/oauth2/default/v1/device/authorize"),
                 bodyPart("client_id", "unit_test_client_id"),
-                bodyPart("scope", "openid%20email%20profile%20offline_access")
+                bodyPart("scope", "openid+email+profile+offline_access")
```
rajdeepnanua-okta: Why change %20 to +? Just curious

okta-direct-auth/src/androidMain/kotlin/com/okta/directauth/DirectAuthenticationFlowBuilder.kt
```diff
@@ -0,0 +1,146 @@
...
+                        builder.apiExecutor,
+                        builder.logger,
+                        builder.clock,
+                        builder.additionalParameter,
+                    )
+                )
+            )
+        }.getOrElse { Result.failure(it) }
+    }
+}
```
rajdeepnanua-okta: Missing newline. Could you add that here? Not sure how spotless didn't catch it

rajdeepnanua-okta (changes requested):
Generally looks good. I have some minor comments. Could you get spotless to apply copyright headers and linting to directauth module? I also think Continuation should be renamed to something specific to our domain to avoid confusion with Kotlin coroutine primitives

auth-foundation/src/main/java/com/okta/authfoundation/GrantType.kt
FeiChen-okta: MFA are different grant types. they are used in different context. Adding them to one would require logic to differentiate the context. for example 

`data class WebAuthn(isMfa: Boolean)`

The usage and documentation will then need to check the context of isMfa for every usage of WebAuthn. It is clearer to specify it as a type.

auth-foundation/src/main/java/com/okta/authfoundation/api/http/log/AuthFoundationLogger.kt
FeiChen-okta: Looks redundant since it is for example my implementation is `val log = AuthFoundationLogger`. The call would be `log.log()`. 
`log.write()` reads better.

okta-direct-auth/src/androidMain/kotlin/com/okta/directauth/DirectAuthenticationFlowBuilder.kt
FeiChen-okta: I will need to setup spotless for this project. I have not yet because of the constant changes. This will be done soon as a clean up.

okta-direct-auth/src/androidMain/kotlin/com/okta/directauth/model/Continuation.kt
```diff
@@ -0,0 +1,83 @@
...
+ * Represents a state in the Direct Authentication flow where further action is required to proceed.
+ *
+ * This is a special type of [DirectAuthenticationState] that indicates the flow is paused,
+ * pending input from the user or another client-side process.
+ *
+ * @param context The [DirectAuthenticationContext] associated with this state.
+ * @param mfaContext An optional context for MFA flows, which may be present if the continuation
+ * is part of a multi-factor sequence.
+ */
+sealed class Continuation(private val context: DirectAuthenticationContext, private val mfaContext: MfaContext? = null) : DirectAuthenticationState {
```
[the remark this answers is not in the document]
FeiChen-okta: Yes agree. The naming is inline with the Swift SDK. It does get confused with coroutine continuation. Will rename this after merging everything. I have a lot of down stream changes that will break if I rename now.

PR #381, opened 2026-04-13
Add cross-platform credential management for KMP (Android + JVM)

FeiChen-okta:
Introduce the KMP credential layer in auth-foundation commonMain, using .kmp sub-packages from the start to cleanly separate from Android-specific implementations:

- credential.kmp.Credential interface and immutable CredentialImpl with Result-returning side-effecting methods and snapshot semantics
- credential.kmp.CredentialDataSource for shared mutable state (token flows, deletion tracking, refresh deduplication)
- credential.kmp.CredentialManager in commonMain holding shared dependencies (client, storage, eventCoordinator, defaultIdStore)
- credential.kmp.TokenStorage and DefaultCredentialIdStore interfaces
- CredentialIdentifier marker interface for cross-platform event polymorphism; event classes use CredentialIdentifier
- SharedFlow-based credential events replacing legacy EventCoordinator callback pattern (fire-and-forget, 64-item buffer)
- JVM CredentialManager wrapper with Consumer-based listener API (CopyOnWriteArrayList, Closeable) and @JvmOverloads
- TokenData, TokenMetadata, TokenInfo additions
- client.kmp.OAuth2Client for cross-platform token operations
- introspectToken, refreshToken, idToken, findAsync on Credential

Android Credential is decoupled from KMP Credential to preserve the original OAuth2ClientResult API. EventCoordinator and RevokeAllException moved to commonMain. Event classes moved from androidMain to commonMain.

Includes 105+ tests across commonTest and jvmTest (including Java tests) covering credential CRUD, immutability, snapshot semantics, events, metadata, token lifecycle, and custom storage.

RESOLVES: OKTA-1115411


auth-foundation/src/commonMain/kotlin/com/okta/authfoundation/credential/events/CredentialCreatedEvent.kt
```diff
@@ -15,17 +15,16 @@
...
- * Emitted via [EventHandler.onEvent] when a [Credential] has been created via a [CredentialDataSource.createCredential] call.
+ * Emitted via [EventHandler.onEvent] when a [CredentialIdentifier] has been created via a createCredential call.
  */
 class CredentialCreatedEvent internal constructor(
     /**
-     * The [Credential] that was created.
+     * The [CredentialIdentifier] that was created.
      */
-    val credential: Credential,
+    val credential: CredentialIdentifier,
```
rajdeepnanua-okta: Rename `credential` to `credentialIdentifier`

auth-foundation/src/commonMain/kotlin/com/okta/authfoundation/credential/kmp/CredentialDataSource.kt
```diff
@@ -0,0 +1,176 @@
...
+                it != null
+            }.filterNotNull()
+    }
+
+    /** Emits a new token value on the shared flow for the given credential [id]. */
+    suspend fun emitToken(
+        id: String,
+        token: TokenData?,
+    ) {
+        val flow = synchronized(tokenFlows) { tokenFlows[id] } ?: return
```
rajdeepnanua-okta, line 69: If this is cross-platform, how is `synchronized` working here? Isn't that a JVM primitive? Also, coroutine-friendly locks should be used in `suspend fun`, such as `Mutex` or `Semaphore`

auth-foundation/src/commonMain/kotlin/com/okta/authfoundation/credential/kmp/CredentialManager.kt
```diff
@@ -0,0 +1,188 @@
...
+            val credential =
+                CredentialImpl(
+                    token = stored,
+                    client = client,
+                    tags = tags,
+                    dataSource = dataSource,
+                    events = _events,
+                    defaultIdStore = defaultIdStore
+                )
+            _events.tryEmit(CredentialCreatedEvent(credential))
```
rajdeepnanua-okta, line 87: Since this is in a `suspend fun`, why not just `emit` instead?

FeiChen-okta: Doesn't really matter in this case because the overflow is set to BufferOverflow.DROP_OLDEST. So even though emit is suspending it actually doesn't suspend. tryEmit makes it explicit that the emitter will not suspended and wait for collectors.

auth-foundation/src/commonMain/kotlin/com/okta/authfoundation/credential/kmp/CredentialDataSource.kt
FeiChen-okta: This is the kotlin synchronized. Though the doc list this as deprecated because of JS is not supported. I revisit once we have a JS target. either wrap this around runBlocking for the JS implementation or use mutex

auth-foundation/src/commonMain/kotlin/com/okta/authfoundation/credential/events/CredentialCreatedEvent.kt
FeiChen-okta: updated

rajdeepnanua-okta (approved):

PR #383, opened 2026-04-13
OAuth2Client Result API, IntrospectInfo model, and cleanup

FeiChen-okta:
OAuth2Client API improvements:
- refreshToken returns Result<TokenInfo> instead of OAuth2ClientResult
- revokeToken returns Result<Unit> instead of OAuth2ClientResult
- introspectToken returns Result<IntrospectInfo> (RFC 7662 model)
- Rename refreshIfExpired/accessTokenIfNotExpired for clarity
- KMP Credential.idToken() returns Result<Jwt>

New IntrospectInfo data class in client.dto with RFC 7662 fields. Remove unused OAuth2ClientResultExtensions.kt.

Android Credential.idToken() reverted to Jwt? return type to avoid breaking Android consumers. Migration docs added to all credential event classes for the Credential -> CredentialIdentifier type change.


auth-foundation/src/jvmMain/kotlin/com/okta/authfoundation/credential/InMemoryTokenStorage.kt
```diff
@@ -1,71 +0,0 @@
...
-
-/**
- * In-memory implementation of [TokenStorage] for JVM applications.
- *
- * **Warning**: Tokens stored in this implementation do **not** persist across process restarts
- * and are **not encrypted**. For production deployments requiring encryption at rest,
- * implement [TokenStorage] with your preferred encryption and configure it during
- * client initialization.
- */
-class InMemoryTokenStorage : TokenStorage {
```
FeiChen-okta, line 30: Delete this placeholder implementation. RoomTokenStorage is now available for jvm. This was added in the previous commit

auth-foundation/src/androidHostTest/kotlin/com/okta/authfoundation/client/DefaultIdTokenValidatorTest.kt
```diff
@@ -124,7 +124,8 @@ class DefaultIdTokenValidatorTest {
     @Test fun testMalformedPayload() {
         val idTokenClaims = IdTokenClaims(issuer = null)
         assertFailsWithMessage(
-            "Expected string value for a non-null key 'iss', got null literal instead at element: \$.iss\n" +
+            "Expected string value for a non-null key 'iss', got null literal instead at path: \$.iss\n" +
+                "Use 'coerceInputValues = true' in 'Json {}' builder to coerce nulls if property has a default value.\n" +
```
FeiChen-okta, line 128: kotlin serialization update changed the error message. Future clean up should just check failure and not match the message.

auth-foundation/src/jvmTest/kotlin/com/okta/authfoundation/credential/InMemoryCommonTokenStorageTest.kt
```diff
@@ -1,140 +0,0 @@
...
-
-import com.okta.authfoundation.credential.kmp.TokenData
-import kotlinx.coroutines.test.runTest
-import kotlin.test.Test
-import kotlin.test.assertEquals
-import kotlin.test.assertNotNull
-import kotlin.test.assertNull
-import kotlin.test.assertTrue
-
-class InMemoryCommonTokenStorageTest {
```
FeiChen-okta, line 26: no longer needed. inmemory token storage is deleted

auth-foundation/src/commonTest/kotlin/com/okta/authfoundation/credential/kmp/InMemoryDefaultCredentialIdStoreTest.kt
```diff
@@ -1,62 +0,0 @@
...
- */
-package com.okta.authfoundation.credential.kmp
-
-import com.okta.authfoundation.credential.InMemoryDefaultCredentialIdStore
-import kotlinx.coroutines.test.runTest
-import kotlin.test.Test
-import kotlin.test.assertEquals
-import kotlin.test.assertNull
-
-class InMemoryDefaultCredentialIdStoreTest {
```
FeiChen-okta, line 24: cleanup. no longer needed. delete this placeholder since it is replaced by roomdefaultcredentila store from previous commit.

auth-foundation/src/androidMain/kotlin/com/okta/authfoundation/credential/Credential.kt
```diff
@@ -516,7 +516,6 @@ class Credential internal constructor(
                 JwtParser(client.configuration.json, client.configuration.computeDispatcher)
             parser.parse(idToken)
         } catch (e: Exception) {
-            // The token was malformed.
             null
```
kavithachandramohan-okta, line 519: do we not log this exception ?

kavithachandramohan-okta:
@FeiChen-okta  these are api changes and all callers must migrate ?

auth-foundation/src/androidMain/kotlin/com/okta/authfoundation/credential/Credential.kt
FeiChen-okta: Nope. The api doc just say return jwt or null. This is unchanged from the current implementation. This is a helper method to return a Jwt class. The use can access the raw string by Credential.token.id

FeiChen-okta:
> @FeiChen-okta these are api changes and all callers must migrate ?

@kavithachandramohan-okta These new APIs are for KMP consumers. The current Android consumers will have min API changes. One minimal change is that Credential now extends CredentialIdentifier. This should not require consumers to change much. 

KMP consumers will first target JVM platform. Once we have a couple of major releases, Android can use the new KMP pattern and the current Android APIs can be deprecated.

rajdeepnanua-okta (approved):

PR #391, opened 2026-05-28
Add KMP TokenExchangeFlow interface, implementation, and Java wrapper

FeiChen-okta:
Implements the Token Exchange Grant (RFC 8693) for KMP targets (Android + JVM). Exchanges an ID token and device secret for a new token set using urn:ietf:params:oauth:grant-type:token-exchange. Includes Java-friendly CompletableFuture wrapper.

RESOLVES: OKTA-1162033


jaredperreault-okta (approved):

khuang-okta (approved):
OptIn is because internalAuthAPI is Internal API and not public?

FeiChen-okta:
> OptIn is because internalAuthAPI is Internal API and not public?
It is public just not guaranteed to be maintained or changed. This is how the previous SDK design was. I'm planning on removing this in the future. It will be internal or public nothing inbetween.

PR #393, opened 2026-06-03
Add KMP SessionTokenFlow interface, implementation, and Java wrapper

FeiChen-okta:
Implements the Session Token flow for KMP targets (Android + JVM). Exchanges a session token for an OAuth2 token set by building an authorization URL (PKCE), making a non-redirecting GET via HttpUtils to capture the Location header, then exchanging the code at the token endpoint. Includes internal HttpUtils helper and Java-friendly CompletableFuture wrapper.

RESOLVES: OKTA-1162035

### DIFF
--- a/oauth2/src/commonMain/kotlin/com/okta/oauth2/internal/HttpUtils.kt
+++ b/oauth2/src/commonMain/kotlin/com/okta/oauth2/internal/HttpUtils.kt
@@ -1,0 +1,90 @@
+/*
+ * Copyright 2022-Present Okta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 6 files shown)


jaredperreault-okta (approved):

oauth2/src/commonMain/kotlin/com/okta/oauth2/internal/HttpUtils.kt
```diff
@@ -0,0 +1,90 @@
...
+                override fun url(): String = url
+
+                override fun contentType(): String = "application/x-www-form-urlencoded"
+
+                override fun formParameters(): Map<String, List<String>> = formParams.mapValues { (_, v) -> listOf(v) }
+            }
+        val response = apiExecutor.execute(request).getOrThrow()
+        val body =
+            response.body?.decodeToString()
+                ?: throw IllegalStateException("Empty response body")
```
khuang-okta, line 60: Could this be decoding error instead of empty response?

FeiChen-okta: no. decoding errors throws exception. that mean body is not empty.

PR #406, opened 2026-07-14
Update web-authentication-ui to use KMP AuthorizationCodeFlow internally. Add KMP OAuth2Client constructor to WebAuthentication; deprecate Android-only path

FeiChen-okta:
- Replace Android-only AuthorizationCodeFlow / RedirectEndSessionFlow imports with the KMP interfaces (com.okta.oauth2.kmp.*) in WebAuthentication
- Use lazy-initialized backing fields so tests can inject stubs before the KMP client creation (which requires HTTPS URLs) is triggered
- Bridge Result<TokenInfo> → OAuth2ClientResult<Token> via a private toToken() helper that builds a Token from the TokenInfo fields + client.configuration
- Fix parameter order in logoutOfBrowser: KMP start() takes (idToken, redirectUrl) whereas the old Android overload took (redirectUrl, idToken)
- Pass uri.toString() to KMP resume() directly instead of the Android compat extension, keeping the call-site free of extra imports
- Update WebAuthenticationTest to stub the KMP flow interfaces instead of making real HTTP calls to the mock server; error tests now expect IllegalStateException (what the KMP impl throws for missing endpoints)
- Update sample app to show a toast on browser sign in errors.

Add KMP OAuth2Client constructor to WebAuthentication; deprecate Android-only path

- Refactor WebAuthentication to a private primary constructor that stores flow factories and a derived OidcConfiguration, decoupling it from any specific client type
- Add preferred constructor taking com.okta.authfoundation.client.kmp.OAuth2Client: creates flows via the KMP companion invoke (no AndroidCompatExtensions bridge), derives OidcConfiguration from KMP config for the Token bridge
- Deprecate the three Android-only constructors (no-arg, OidcConfiguration, OAuth2Client) with ReplaceWith pointing to the KMP constructor
- Update API dump to include the new KMP constructor alongside the deprecated ones
- Show real exception message in BrowserViewModel error state and add a toast in BrowserFragment so login errors are visible in the sample app
- Fix app content rendering under status bar on Android 15
- Update sample app to use KMP OAuth2Client constructor for WebAuthentication

RESOLVES: OKTA-1163828

### DIFF
--- a/web-authentication-ui/src/main/java/com/okta/webauthenticationui/WebAuthentication.kt
+++ b/web-authentication-ui/src/main/java/com/okta/webauthenticationui/WebAuthentication.kt
@@ -18,16 +18,20 @@ package com.okta.webauthenticationui
 import android.app.Activity
 import android.content.Context
 import androidx.annotation.VisibleForTesting
-import com.okta.authfoundation.client.OAuth2Client
 import com.okta.authfoundation.client.OAuth2ClientResult
 import com.okta.authfoundation.client.OidcConfiguration
+import com.okta.authfoundation.client.TokenInfo
 import com.okta.authfoundation.client.internal.SdkVersionsRegistry
+import com.okta.authfoundation.client.kmp.OAuth2Client
 import com.okta.authfoundation.credential.Token
-import com.okta.oauth2.AuthorizationCodeFlow
-import com.okta.oauth2.RedirectEndSessionFlow
+import com.okta.authfoundation.events.EventCoordinator
+import com.okta.oauth2.kmp.AuthorizationCodeFlow
+import com.okta.oauth2.kmp.RedirectEndSessionFlow
 import com.okta.webauthenticationui.BuildConfig.SDK_VERSION
 import com.okta.webauthenticationui.events.CustomizeBrowserEvent
 import com.okta.webauthenticationui.events.CustomizeCustomTabsEvent
+import okhttp3.HttpUrl.Companion.toHttpUrl
+import com.okta.authfoundation.client.OAuth2Client as AndroidOAuth2Client
 
 @Deprecated(
     message = "Renamed to WebAuthentication",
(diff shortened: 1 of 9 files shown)


web-authentication-ui/src/main/java/com/okta/webauthenticationui/WebAuthentication.kt
```diff
@@ -45,25 +49,63 @@ typealias WebAuthenticationClient = WebAuthentication
...
+     * This is the preferred constructor. The KMP client is used directly to create the underlying
+     * [AuthorizationCodeFlow] and [RedirectEndSessionFlow].
+     *
+     * @param oAuth2Client the KMP [OAuth2Client] to use for authorization requests.
+     * @param webAuthenticationProvider the [WebAuthenticationProvider] which will be used to show
+     * the UI when performing the redirect flows.
+     */
+    constructor(
+        oAuth2Client: OAuth2Client,
+        webAuthenticationProvider: WebAuthenticationProvider = DefaultWebAuthenticationProvider(EventCoordinator(emptyList())),
```
khuang-okta, line 77: Why emptyList and not keep client.configuration.eventCoordinator?

FeiChen-okta: Because the DefaultWebAuthenticationProvider is from the SDK and does not know the apps usage of the events. The SDK can't provide event handling.

FeiChen-okta: also if you look at `client.configuration.eventCoordinator` it is a empty list 🤣

FeiChen-okta: also the default client.configuration.eventCoordinator is android only. this change is using multi platform dependencies

khuang-okta (approved):

PR #412, opened 2026-07-29
Deprecate Android-only auth-foundation/oauth2 APIs in favor of KMP

FeiChen-okta:
Adds an Android Keystore/biometric-capable KMP TokenEncryptionHandler (the KMP credential storage stack previously only worked on JVM), adds List<String> scope overloads alongside the existing String ones for consistency, deprecates the Android-only OAuth2Client/OidcConfiguration/ Credential/CredentialDataSource/TokenStorage/RoomTokenStorage/Token and the six com.okta.oauth2.* flow classes in favor of their KMP equivalents, migrates the app sample to the KMP variant end-to-end, and documents the migration path in README.md.

RESOLVES: OKTA-1209077

### DIFF
--- a/oauth2/src/commonMain/kotlin/com/okta/oauth2/kmp/AuthorizationCodeFlow.kt
+++ b/oauth2/src/commonMain/kotlin/com/okta/oauth2/kmp/AuthorizationCodeFlow.kt
@@ -59,6 +59,20 @@ interface AuthorizationCodeFlow {
     ): Result<AuthorizationCodeFlowContext>
 
     /**
+     * Initiates the Authorization Code redirect flow.
+     *
+     * @param redirectUrl the registered redirect URI for this client.
+     * @param extraRequestParameters additional query parameters for the authorization endpoint.
+     * @param scope the scopes to request.
+     * @return a [Result] containing an [AuthorizationCodeFlowContext] with the authorization URL and state.
+     */
+    suspend fun start(
+        redirectUrl: String,
+        extraRequestParameters: Map<String, String> = emptyMap(),
+        scope: List<String>,
+    ): Result<AuthorizationCodeFlowContext> = start(redirectUrl, extraRequestParameters, scope.joinToString(" "))
+
+    /**
      * Completes the Authorization Code redirect flow by exchanging the authorization code for tokens.
      *
      * @param uri the redirect URI received from the browser after authorization.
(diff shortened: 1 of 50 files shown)


oauth2/src/commonMain/kotlin/com/okta/oauth2/kmp/AuthorizationCodeFlow.kt
```diff
@@ -58,6 +58,20 @@ interface AuthorizationCodeFlow {
...
+     * @param redirectUrl the registered redirect URI for this client.
+     * @param extraRequestParameters additional query parameters for the authorization endpoint.
+     * @param scope the scopes to request.
+     * @return a [Result] containing an [AuthorizationCodeFlowContext] with the authorization URL and state.
+     */
+    suspend fun start(
+        redirectUrl: String,
+        extraRequestParameters: Map<String, String> = emptyMap(),
+        scope: List<String>,
+    ): Result<AuthorizationCodeFlowContext> = start(redirectUrl, extraRequestParameters, scope.joinToString(" "))
```
BenjaminTruong-okta, line 73: when `scope` is an empty list, should we pass `null` so that the default scope fallback triggers?

https://github.com/okta/okta-mobile-kotlin/blob/fc-OKTA-1209077-deprecate-android-only-api/oauth2/src/commonMain/kotlin/com/okta/oauth2/kmp/AuthorizationCodeFlowImpl.kt#L58

(same on the other flow interfaces)

FeiChen-okta: its not a empty list. 
`scope : List<String>` is a required parameter. Callers need to set the scope. they can put in emptyList. that is up to them and that will be honored.

BenjaminTruong-okta (approved):

oauth2/src/commonMain/kotlin/com/okta/oauth2/kmp/AuthorizationCodeFlow.kt
FeiChen-okta: one more clarification is that the new kmp API will discourage hidden defaults. This hides functionality and unexpected behavior. such as the previous API where you can call
`start(redirectUrl)` then callers will not know what the scope is and will need to look for the defaults in a separate object. Defaults should be in method signatures that is easily inspectable.

PR #414, opened 2026-07-30
Support aud claim as JSON array in ID token validation

FeiChen-okta:
Per RFC 7519 §4.1.3 / OIDC Core 1.0 §3.1.3.7, the `aud` claim may be either a single string or a JSON array of strings. Add an AudienceSerializer to normalize both forms to a List<String>, and update audience validation in the Android and common DefaultIdTokenValidator implementations to check that the client ID is contained in the audience list rather than requiring an exact string match.


BenjaminTruong-okta (approved):
lgtm

PR #418, opened 2026-08-05
Add closable

FeiChen-okta:
The Java continuation wrappers need a closable to close the
internal scopes. Update the samples on how this is used.
Add back the KeygenParameterSpec workaround on Android.
In DirectAuthContinuation fix a issue where the flow was not updated.

PR #419, opened 2026-08-05
Fix and expand KDoc across all Dokka-published modules

FeiChen-okta:
Applies fixes from a full KDoc audit of the six modules published to the https://okta.github.io/okta-mobile-kotlin reference site (auth-foundation, oauth2, web-authentication-ui, legacy-token-migration, okta-idx-kotlin, okta-direct-auth), scoped to their public (non-internal) API surface.

Inaccurate KDoc corrected (verified against current behavior, not just reworded):
- auth-foundation: KMP Credential/CredentialImpl no longer claim the deprecated Android Credential implements them (it doesn't - shares only CredentialIdentifier); fixed an ambiguous same-package [Credential] link that silently resolved to the wrong class; fixed a wrong FQN on DefaultCredentialIdStore's linked type.
- oauth2: TokenExchangeFlow's audience no longer claims a fabricated "api://default" fallback the KMP path doesn't have; ResourceOwnerFlow's example no longer instantiates an internal class external readers can't see; fixed five JVM-wrapper @ReplaceWith expressions that referenced a non-null-safe scope.split(" ") against a nullable parameter (wouldn't compile if applied); removed redundant/invalid lowercase @deprecated tags left over from a Javadoc convention on six deprecated Android flow classes.
- legacy-token-migration: migrate()'s docs no longer describe a "caller stores the returned Token" contract the code doesn't implement (it stores the credential internally); SuccessfullyMigrated no longer references a Credential parameter that doesn't exist on migrate(); filled in a literally empty KDoc block on Result.Error.exception; fixed the same stale API references in migrate.md.
- okta-idx-kotlin: IdxResponse.isLoginSuccessful no longer points to a nonexistent exchangeCode method (real one is exchangeInteractionCodeForTokens); documented IdxMessage.localizationKey's nullability instead of implying it's always present; fixed copy-pasted "WebAuthn activation" wording on the authentication (assertion) capability and a truncated @throws sentence.
- okta-direct-auth: DirectAuthenticationFlow no longer points to a nonexistent DirectAuthenticationFlow.create factory; corrected a commonMain logger doc that claimed Android Logcat behavior unconditionally on a KMP (Android+JVM) type; corrected the builder's description of how it's actually used (configured via create()'s trailing lambda, not a held builder instance).

Missing KDoc added on public declarations that rendered blank on the doc site (TokenInfo, Event, InternalAuthFoundationApi, six ForegroundUIEvent objects, IdxUser.Profile, AuthenticatorEnrollment properties, and others), plus @return/@throws/@param coverage added to primary entry points that previously documented only the success path (AuthorizationCodeFlow.resume, WebAuthentication.login/logoutOfBrowser, InteractionCodeFlow.proceed/ exchangeInteractionCodeForTokens/evaluateRedirectUri/resume, DirectAuthenticationFlow.start, WebAuthenticationProvider.launch).

Added guidance where docs were accurate but didn't orient a new reader: call-sequence notes on AuthorizationCodeFlow and InteractionCodeFlow, the WebAuthn ceremony handshake on WebAuthnCeremonyHandler, cross-pointers between the Kotlin-native and Java-CompletableFuture API pairs in oauth2 and okta-direct-auth, and threading notes on JVM listener callbacks.

Deliberately not included (separate follow-up, since they'd need ABI re-dumps and/or code changes rather than doc text):
- Making okta-direct-auth's com.okta.directauth.http request/error types and okta-idx-kotlin's DeviceTokenProvider internal - they're public with no KDoc and aren't caught by the existing *.internal.* Dokka suppression rule.
- auth-foundation KMP Credential.refreshToken(extraRequestParameters)'s default implementation contradicts its own "bypasses the coalescing orchestrator" KDoc (falls back to refreshToken(), which doesn't).
- okta-direct-auth's AuthorizationPending.timestamp KDoc says milliseconds; confirmed via TokenStepHandler that it's actually populated in seconds.

Also standardizes module README installation instructions and brings the CHANGELOG in line with what's actually shipped:

- Installation: oauth2, okta-direct-auth, and okta-idx-kotlin now install via the BOM (matching auth-foundation/web-authentication-ui) instead of a pinned per-artifact version that could drift from BOM_VERSION. oauth2 and okta-direct-auth's snippets now also list the auth-foundation dependency they actually require at compile time (oauth2 exposes it as `api`, but okta-direct-auth only as `implementation`, so callers touching its auth-foundation-typed return values need to add it themselves). okta-idx-kotlin's snippet modernized from old Groovy single-quote syntax.
- okta-direct-auth/README.md: bumped stale "Current Version: 0.0.1" to 1.0.0; fixed Kotlin/Java examples using renamed params (loginHint/ primaryFactor/secondaryFactor/challengeTypesSupported) and a non-existent DirectAuthenticationError.OAuth2Error type path (real one is nested HttpError.Oauth2Error); replaced a stub Java error-handling section with real subtype examples; corrected the Requirements section (removed a redundant bullet describing the module rather than a real requirement, and clarified "Android API 26+ or JVM (Java 11+)" means whichever platform you target, not both).
- web-authentication-ui/README.md: added a Usage section demonstrating login()/logoutOfBrowser() (previously covered only setup/troubleshooting, with no working example).
- CHANGELOG.md: added the missing entry for the com.okta.directauth.jvm continuation wrappers implementing Closeable (shipped in #418 without a changelog entry). Removed three "Fixed" entries for bugs that were introduced and fixed within this same unreleased cycle and therefore never shipped in a release (a Credential.refreshToken(extraRequestParameters) NotImplementedError fix, an Android Keystore RSA re-import regression in TokenEncryptionHandler/AndroidTokenEncryptionHandler, and a DirectAuthContinuation.WebAuthn.proceed(handler) StateFlow bug) — verified each remaining "Fixed" entry against its module's last released tag to confirm it documents a real, previously-shipped bug.

- okta-direct-auth/CHANGELOG.md was stale (only had the 0.0.1 alpha entry) despite its own README linking to it as the place for "release history and migration guides." Added a [1.0.0] - Unreleased section mirroring root CHANGELOG.md's content for this module.
- Root CHANGELOG.md's `okta-direct-auth 1.0.0` section removed in favor of that per-module CHANGELOG.md, matching how okta-idx-kotlin's own CHANGELOG.md is already the sole source for its changes — avoids two copies of the same release notes drifting apart. Updated the root Unreleased intro accordingly to point at okta-direct-auth/CHANGELOG.md.

- Release checklist step 3 expanded with a concrete template for how the "Unreleased" section should be split into dated per-module entries at release time, including the [Commits](...) compare-link format (matching older dated entries in this file). Templated auth-foundation's link from its real auth-foundation@2.0.5 tag; oauth2/web-authentication-ui have no prior per-module tag to compare from (this is their first independent release), so their templates compare from commit e70ffa5c instead, the last commit before either module's version left 2.0.4.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,11 @@
 ## Unreleased
 
 Per-module versions are now tracked and released independently. The sections below cover
-`auth-foundation`, `oauth2`, `web-authentication-ui`, and `okta-direct-auth` changes accumulated
-since each module's last release. `okta-idx-kotlin` (currently 3.1.1) is unaffected and not
-included here.
+`auth-foundation`, `oauth2`, and `web-authentication-ui` changes accumulated since each module's
+last release. `okta-direct-auth` and `okta-idx-kotlin` each maintain their own `CHANGELOG.md`
+(`okta-direct-auth/CHANGELOG.md`, `okta-idx-kotlin/CHANGELOG.md`) and are not duplicated here —
+see `okta-direct-auth/CHANGELOG.md` for its upcoming 1.0.0 changes. `okta-idx-kotlin` (currently
+3.1.1) is additionally unaffected by this release wave.
 
 ### Release checklist (next steps)
 
@@ -17,7 +19,35 @@ included here.
    CLAUDE.md's API Compatibility section).
 3. Move this "Unreleased" section into dated, per-module entries once each release is actually
    tagged and published (see the existing per-module tags: `auth-foundation@2.0.5`,
-   `okta-direct-auth@0.0.1`, `okta-idx-kotlin@3.1.1`).
+   `okta-direct-auth@0.0.1`, `okta-idx-kotlin@3.1.1`). Replace each `### <module> 3.0.0` heading
+   with a top-level, dated `## <module> 3.0.0 - <release date>` heading plus a commit-compare
+   link, matching the `[Commits](...)` links on older entries further down this file:
+
+   ```
+   ## auth-foundation 3.0.0 - <release date>
+
+   [Commits](https://github.com/okta/okta-mobile-kotlin/compare/auth-foundation@2.0.5...auth-foundation@3.0.0)
+
+   #### Breaking changes
+   ...
+   ```
+
+   `oauth2` and `web-authentication-ui` are being tagged independently for the first time with
+   this release, so there's no prior `oauth2@2.0.4`/`web-authentication-ui@2.0.4` tag to compare
+   from — use commit `e70ffa5c` (the last commit before either module's version left 2.0.4)
+   instead:
+
+   ```
+   ## oauth2 3.0.0 - <release date>
+
+   [Commits](https://github.com/okta/okta-mobile-kotlin/compare/e70ffa5c...oauth2@3.0.0)
+   ```
+
+   ```
+   ## web-authentication-ui 3.0.0 - <release date>
+
+   [Commits](https://github.com/okta/okta-mobile-kotlin/compare/e70ffa5c...web-authentication-ui@3.0.0)
+   ```
 
 ### auth-foundation 3.0.0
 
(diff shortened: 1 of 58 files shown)


CHANGELOG.md
```diff
@@ -17,7 +19,35 @@ included here.
    CLAUDE.md's API Compatibility section).
 3. Move this "Unreleased" section into dated, per-module entries once each release is actually
    tagged and published (see the existing per-module tags: `auth-foundation@2.0.5`,
-   `okta-direct-auth@0.0.1`, `okta-idx-kotlin@3.1.1`).
+   `okta-direct-auth@0.0.1`, `okta-idx-kotlin@3.1.1`). Replace each `### <module> 3.0.0` heading
```
FeiChen-okta, line 22: this section will be updated with another PR once the artifacts are published.

khuang-okta (approved):

PR #422, opened 2026-08-10
feat: Add support for ephemeral browsing in Custom Tabs

kallurajput80:
## Summary

This change adds optional support for **ephemeral browsing** in the web authentication flow.

## Issue

When a customer switches between ** type 1 account to type 2** or ** type 2 account to type 1** flows, the login screen can reuse an already active browser session instead of reflecting the updated user context. Because of that, it's picking previously logged-in information instead of forcing a fresh authentication experience.

## Solution

To address this, this PR introduces an **Ephemeral Browsing** option in the web authentication flow.

When enabled, the SDK will request an ephemeral browser session (when supported by the selected Custom Tabs browser). This helps prevent reuse of stale browser session data and ensures the authentication flow reflects the latest user context.

### What changed

- Added an  `enableEphemeralBrowsing` option through the web authentication flow APIs.
- Propagated the ephemeral browsing flag through:
  - `WebAuthentication`
  - `RedirectCoordinator`
  - `WebAuthenticationProvider`
  - `DefaultWebAuthenticationProvider`
- Enabled ephemeral browsing only when the selected browser supports it.
- Updated unit tests to cover the new API behavior and event flow expectations.

## Backward compatibility

This feature is **optional** and defaults to `false`.

That means:
- existing consumers of the library are **not affected**
- current integrations continue to behave exactly as before unless they explicitly opt in
- consumers who do not need this behavior do not need to make any code changes

## Benefits

- avoids stale session reuse across user/dealer switching scenarios
- improves correctness of authentication context
- preserves existing behavior for all current SDK consumers

## Testing

- Updated `web-authentication-ui` unit tests
- Verified affected build/test flows pass successfully

### DIFF
--- a/web-authentication-ui/src/main/java/com/okta/webauthenticationui/DefaultWebAuthenticationProvider.kt
+++ b/web-authentication-ui/src/main/java/com/okta/webauthenticationui/DefaultWebAuthenticationProvider.kt
@@ -22,6 +22,7 @@ import android.content.pm.PackageManager
 import android.os.Build
 import android.os.Bundle
 import android.provider.Browser
+import androidx.browser.customtabs.CustomTabsClient
 import androidx.browser.customtabs.CustomTabsIntent
 import androidx.browser.customtabs.CustomTabsService
 import androidx.core.net.toUri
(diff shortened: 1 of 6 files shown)


FeiChen-okta:
Hello @kallurajput80 Thank you for your contribution. The latest SDK exposes the CustomTabsIntent.Builder for applications to customize CCT. This is not just limited to ephemeral browser. I have an example for using ephemeral browsing: https://github.com/okta/okta-mobile-kotlin/pull/424

PR #425, opened 2026-08-12
Remove corelibrary desugaring

FeiChen-okta:
Min SDK 26+ does not require desugaring.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -50,7 +50,6 @@ android {
     }
 
     compileOptions {
-        isCoreLibraryDesugaringEnabled = true
         sourceCompatibility = SOURCE_COMPATIBILITY
         targetCompatibility = TARGET_COMPATIBILITY
     }
@@ -78,8 +77,6 @@ kotlin {
 }
 
 dependencies {
-    coreLibraryDesugaring(libs.core.library.desugaring)
-
     implementation(project(":web-authentication-ui"))
 
     implementation(libs.kotlin.stdlib)
--- a/auth-foundation/build.gradle.kts
+++ b/auth-foundation/build.gradle.kts
@@ -215,5 +215,4 @@ room {
 dependencies {
     add("kspAndroid", libs.room.compiler)
     add("kspJvm", libs.room.compiler)
-    coreLibraryDesugaring(libs.core.library.desugaring)
 }
--- a/dynamic-app/build.gradle.kts
+++ b/dynamic-app/build.gradle.kts
@@ -62,7 +62,6 @@ android {
     }
 
     compileOptions {
-        isCoreLibraryDesugaringEnabled = true
         sourceCompatibility = SOURCE_COMPATIBILITY
         targetCompatibility = TARGET_COMPATIBILITY
     }
@@ -92,7 +91,6 @@ kotlin {
 }
 
 dependencies {
-    coreLibraryDesugaring(libs.core.library.desugaring)
     implementation(project(":okta-idx-kotlin"))
 
     implementation(libs.kotlin.stdlib)
--- a/legacy-token-migration-sample/build.gradle.kts
+++ b/legacy-token-migration-sample/build.gradle.kts
@@ -54,7 +54,6 @@ android {
     }
 
     compileOptions {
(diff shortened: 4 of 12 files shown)


TanviMandanaka-okta (approved):

harikun-okta (commented):
Claude flagged two issues. Can you check if they are valid

**1. Correctness issue — incomplete removal**

Two Android modules still reference desugaring and were not touched by this PR:
  - okta-direct-auth-android-app/build.gradle.kts:41,51 — isCoreLibraryDesugaringEnabled = true and coreLibraryDesugaring(libs.core.library.desugaring), minSdk = 28
  - okta-direct-auth-shared/build.gradle.kts:133 — coreLibraryDesugaring(libs.core.library.desugaring), minSdk = 28

  Both have minSdk ≥ 26, so by the PR's own stated rationale ("Min SDK 26+ does not require desugaring") these should have been cleaned up too. This looks like an oversight rather than an
  intentional exclusion — worth confirming with the author whether these were skipped deliberately (e.g. still transitioning) or missed.

  Because these two modules still use libs.core.library.desugaring, the version catalog entries (gradle/libs.versions.toml:23,109) correctly remain — no dead-catalog-entry issue, but only as a
  side effect of the incomplete removal above.

**2.  Risk: desugaring isn't purely a minSdk<26 backstop**

  Core library desugaring also backports Java APIs added after API 26 (e.g., newer java.time, java.util.stream, NIO2, sequenced-collection methods, depending on the desugar_jdk_libs version in
  use). "minSdk 26+" alone doesn't guarantee every Java 8+/9+/11+ API used in the codebase is natively available at API 26 — it depends on which specific APIs the code calls. Since there are
  no source changes in this diff, if any module relies on such backported APIs, this would surface as a build/lint failure (missing API) rather than a silent runtime issue — which lines up
  with the failing CI check below.

FeiChen-okta:
> Claude flagged two issues. Can you check if they are valid
> 
> **1. Correctness issue — incomplete removal**
> 
> Two Android modules still reference desugaring and were not touched by this PR:
> 
> * [okta-direct-auth-android-app](https://oktainc.atlassian.net/browse/OKTA-direct-auth-android-app)/build.gradle.kts:41,51 — isCoreLibraryDesugaringEnabled = true and coreLibraryDesugaring(libs.core.library.desugaring), minSdk = 28
> * [okta-direct-auth-shared](https://oktainc.atlassian.net/browse/OKTA-direct-auth-shared)/build.gradle.kts:133 — coreLibraryDesugaring(libs.core.library.desugaring), minSdk = 28
> 
> Both have minSdk ≥ 26, so by the PR's own stated rationale ("Min SDK 26+ does not require desugaring") these should have been cleaned up too. This looks like an oversight rather than an intentional exclusion — worth confirming with the author whether these were skipped deliberately (e.g. still transitioning) or missed.
> 
> Because these two modules still use libs.core.library.desugaring, the version catalog entries (gradle/libs.versions.toml:23,109) correctly remain — no dead-catalog-entry issue, but only as a side effect of the incomplete removal above.
> 
> **2. Risk: desugaring isn't purely a minSdk<26 backstop**
> 
> Core library desugaring also backports Java APIs added after API 26 (e.g., newer java.time, java.util.stream, NIO2, sequenced-collection methods, depending on the desugar_jdk_libs version in use). "minSdk 26+" alone doesn't guarantee every Java 8+/9+/11+ API used in the codebase is natively available at API 26 — it depends on which specific APIs the code calls. Since there are no source changes in this diff, if any module relies on such backported APIs, this would surface as a build/lint failure (missing API) rather than a silent runtime issue — which lines up with the failing CI check below.

1. This is on purpose because those are sample apps. Even though min SDK = 28, the samples can use dependencies that require desugaring.

2. The failure is CircleCI infra: 
Downloading https://services.gradle.org/distributions/gradle-9.6.1-bin.zip
Attempt 1/1 failed. Reason: Unexpected end of file from server
Exception in thread "main" java.net.SocketException: Unexpected end of file from server.
Not caused by removal of desugaring. 
The SDK modules is not using any of the later Java11+ API such as nio, list, set

harikun-okta (approved):

PR #435, opened 2026-09-02
Implement the Cross App Access two-step exchange

FeiChen-okta:
CrossAppAccessFlow.create builds (or adopts) the target OAuth2Client and validates it can authenticate before any network request. start() exchanges a subject assertion for an ID-JAG at the IdP authorization server; redeem() presents it at the target for a resource access token; exchange() composes the two.

Wire-level tests for start()/redeem()/exchange() covering all three subject forms, endpoint override precedence, re-redemption, event discrimination, independent target client identity, failure attribution, client-assertion rebinding, concurrency, conditional parameters, and audience resolution — plus interface-level pre-flight and redaction tests.

RESOLVES: OKTA-1258536

### DIFF
--- a/oauth2/src/commonMain/kotlin/com/okta/oauth2/kmp/CrossAppAccessFlowImpl.kt
+++ b/oauth2/src/commonMain/kotlin/com/okta/oauth2/kmp/CrossAppAccessFlowImpl.kt
@@ -122,6 +122,11 @@ internal class CrossAppAccessFlowImpl(
             runCatching {
                 val targetClient = buildTargetClient(idpClient, target)
                 val configuration = targetClient.configuration
+                require(configuration.issuerUrl.trimEnd('/') != idpClient.configuration.issuerUrl.trimEnd('/')) {
+                    "The target resource authorization server (${configuration.issuerUrl}) resolves to the same issuer as " +
+                        "the IdP authorization server (${idpClient.configuration.issuerUrl}). Cross App Access requires " +
+                        "two distinct authorization servers."
+                }
                 require(configuration.clientSecret.isNotBlank() || configuration.clientAssertionProvider != null) {
                     "The target resource authorization server has no client secret or client assertion " +
                         "provider configured. Cross App Access requires a confidential client at both steps."
--- a/oauth2/src/commonMain/kotlin/com/okta/oauth2/kmp/CrossAppAccessFlowImpl.kt
+++ b/oauth2/src/commonMain/kotlin/com/okta/oauth2/kmp/CrossAppAccessFlowImpl.kt
@@ -15,6 +15,8 @@
  */
 package com.okta.oauth2.kmp
 
+import com.okta.authfoundation.InternalAuthFoundationApi
+import com.okta.authfoundation.client.OAuth2ClientBuilder
 import com.okta.authfoundation.client.TokenInfo
 import com.okta.authfoundation.client.kmp.OAuth2Client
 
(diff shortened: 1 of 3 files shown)


oauth2/src/commonMain/kotlin/com/okta/oauth2/kmp/CrossAppAccessFlowImpl.kt
```diff
@@ -23,29 +25,158 @@ import com.okta.authfoundation.client.kmp.OAuth2Client
...
+    ): Result<IdJagAssertion> =
+        runCatching {
+            // Client-side narrowing, not a spec requirement: draft §4.3 marks `scope` OPTIONAL,
+            // but Okta's org authorization server returns invalid_scope for a scope-less
+            // exchange.
+            val resolvedScope = scope ?: defaultScope
+            require(!resolvedScope.isNullOrEmpty()) {
+                "No scope was requested for this Cross App Access exchange. Supply scope on " +
+                    "CrossAppAccessTarget or on start() — the IdP authorization server rejects a " +
+                    "scope-less request for this grant."
```
AlexNachbaur, line 50: Hmm this is weird, I didn't see that in my tests. This might be a gap in the python SDK too then.

oauth2/src/commonMain/kotlin/com/okta/oauth2/kmp/CrossAppAccessFlowImpl.kt
```diff
@@ -23,29 +25,158 @@ import com.okta.authfoundation.client.kmp.OAuth2Client
...
-    override fun toString(): String = TODO("implemented in the https://oktainc.atlassian.net/browse/OKTA-1258536")
+    override fun toString(): String = "CrossAppAccessFlow(idp=${idpClient.configuration.issuerUrl}, target=${targetClient.configuration.issuerUrl})"
 
     companion object {
+        /**
+         * A single, never-transmitted scope value used only to satisfy [OAuth2ClientBuilder]'s
+         * non-empty-scope requirement when a target names no scopes of its own. [start] resolves
+         * its outgoing `scope` from the [CrossAppAccessTarget] descriptor directly and never reads
+         * the built target client's configured default scope, so this value never reaches the wire.
+         */
```
AlexNachbaur, line 115: Instead of using a placeholder scope, shouldn't we instead update OAuth2ClientBuilder to accept an empty or null scope list?  A lack of scopes is fully permitted within the OAuth2 specifications, and should be generally supported.

FeiChen-okta: Agree that the OAuth specification permits empty scopes, but Okta's server requires them and returns invalid_scope without one. Enforcing a non-empty scope in OAuth2ClientBuilder acts as a fail-fast check to prevent developers from making errors that are easily avoidable ahead of time. This inert scope is odd when i added it but I want to keep OAuth2ClientBuilder as is so the targetClient creation is never exposed. It is created from a CrossAppAccessTarget data class, so developers will never see this scope and it is never used.

oauth2/src/commonMain/kotlin/com/okta/oauth2/kmp/CrossAppAccessFlowImpl.kt
FeiChen-okta: Yes. this is not part of the RFC but Okta is enforcing it. This maybe relaxed now, I can this this once implementation is completed.

TanviMandanaka-okta, line 50: the docs promise it rejects a target pointing at the same server as the IdP, but the code never checks

TanviMandanaka-okta (approved):

oauth2/src/commonMain/kotlin/com/okta/oauth2/kmp/CrossAppAccessFlowImpl.kt
FeiChen-okta: Good catch. Fixed

PR #438, opened 2026-09-10
Implement Java wrapper for Cross App Access

FeiChen-okta:
CrossAppAccessFlow (jvm) delegates start/redeem/exchange to CompletableFuture via the wrapper's coroutine scope, create() returns AuthFoundationResult rather than throwing, and CrossAppAccessTargetBuilder (jvm) delegates to the commonMain factories through chaining setters. All 16 Java-source tests pass, confirming the surface is genuinely Java-idiomatic (compiler-enforced, not just documented).

RESOLVES: OKTA-1258538

### DIFF
--- a/oauth2/src/jvmMain/kotlin/com/okta/oauth2/kmp/jvm/CrossAppAccessFlow.kt
+++ b/oauth2/src/jvmMain/kotlin/com/okta/oauth2/kmp/jvm/CrossAppAccessFlow.kt
@@ -1,0 +1,175 @@
+/*
+ * Copyright 2022-Present Okta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.okta.oauth2.kmp.jvm
+
+import com.okta.authfoundation.client.TokenInfo
+import com.okta.authfoundation.client.jvm.AuthFoundationResult
+import com.okta.oauth2.kmp.crossAppAccessSubject
+import com.okta.oauth2.kmp.crossAppAccessToken
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.future.future
+import java.io.Closeable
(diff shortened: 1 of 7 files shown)


oauth2/src/jvmMain/kotlin/com/okta/oauth2/kmp/jvm/CrossAppAccessFlow.kt
```diff
@@ -0,0 +1,175 @@
...
+         * @return a [CompletableFuture] that completes with the [KotlinSubjectAssertion] on
+         *   success, or completes exceptionally with an `IllegalStateException` naming the
+         *   missing token on failure.
+         */
+        @JvmStatic
+        @JvmOverloads
+        fun crossAppAccessSubject(
+            credential: KmpCredential,
+            type: KotlinSubjectAssertion.Type = KotlinSubjectAssertion.Type.ID_TOKEN,
+        ): CompletableFuture<KotlinSubjectAssertion> = CoroutineScope(Dispatchers.Default).future { credential.crossAppAccessSubject(type).getOrThrow() }
```
harikunwar-okta, line 148: From Claude:

Missing SupervisorJob in static companion scope constructors

  CrossAppAccessFlow.kt lines 147 and 162:

  // Current — no SupervisorJob
  fun crossAppAccessSubject(...): CompletableFuture<KotlinSubjectAssertion> =
      CoroutineScope(Dispatchers.Default).future { ... }

  fun crossAppAccessToken(...): CompletableFuture<TokenInfo> =
      CoroutineScope(Dispatchers.Default).future { ... }

  When the coroutine fails (e.g. getOrThrow() throws), the exception propagates to the scope's Job, cancelling it. With no SupervisorJob, that scope — and any other coroutines sharing it —
  gets cancelled. For one-shot ephemeral scopes this is harmless today, but it's inconsistent with the instance scope that correctly uses SupervisorJob():

  private val coroutineScope = CoroutineScope(SupervisorJob() + Dispatchers.Default)

  Fix: Use the same pattern for the static helpers:
  
  fun crossAppAccessSubject(...): CompletableFuture<KotlinSubjectAssertion> =
      CoroutineScope(SupervisorJob() + Dispatchers.Default).future { ... }

  fun crossAppAccessToken(...): CompletableFuture<TokenInfo> =
      CoroutineScope(SupervisorJob() + Dispatchers.Default).future { ... }

FeiChen-okta: @harikunwar-okta 
This is intentional. It is in the KDocs
```
Runs on its own short-lived coroutine scope; unlike the instance methods above, it is unaffected by [close].
```
It documents that these static helpers deliberately run on their own independent scope specifically because they're not tied to any CrossAppAccessFlow instance — there's no close() to be affected by, since the method takes a existing Credential, not this. These are JVM static methods used by Java. SupervisorJob() vs Job() only matters when a scope has multiple children that shouldn't cancel each other. These static-helper call creates a fresh scope with exactly one child coroutine, used once, then discarded. There's no sibling ever attached, so swapping in SupervisorJob() would be a no-op — it wouldn't change behavior.

helenchen-okta (approved):

